PR #17, opened 2022-09-10
Add Semgrep as a new linter

reedloden:
https://semgrep.dev/docs/semgrep-ci/overview/


wadells:
Lets pick this back up if/when it is a priority.

PR #24, opened 2022-10-03
Limit classify changes to Teleport repo

jimbishopp:
The `classifyChanges` function determines whether a file path includes `/docs` and chooses different reviewers. This should only apply to the Teleport repo.

In order to support a repo comparison, the repo and teams constants were moved from the review package to the env package.


gozer (approved):

r0mant (approved):
You could've probably achieved the same without any code changes and just adding folks who should review docs to "docs reviewers" in the Cloud-specific json map? But lgtm either way.

jimbishopp:
@r0mant Thanks for reviewing. Docs in the cloud repo are internal runbooks and things - no need for different reviewers like in Teleport. For path-based reviewers, we're using the preferred reviewers functionality you created. I think this specific docs logic should be stripped out and Teleport should use preferred reviewers so we have one way to set reviewers based on path.

PR #27, opened 2022-10-04
bot: Only owners of the current repo are candidates

gozer:
Currently, the `owner: true` flag is per-user, ignoring whatever team that user belongs to. This causes owners to be added to the required reviewers list across repos, it shouldn't.

### DIFF
--- a/bot/internal/review/review.go
+++ b/bot/internal/review/review.go
@@ -399,7 +399,17 @@ func getReviewerSets(e *env.Environment, team string, reviewers map[string]Revie
 			continue
 		}
 
-		if v.Owner {
+		team := v.Team
+
+		switch e.Repository {
+		case env.TeleportRepo:
+			team = env.CoreTeam
+		case env.CloudRepo:
+			team = env.CloudTeam
+		}
+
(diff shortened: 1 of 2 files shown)


bot/internal/review/review.go
```diff
@@ -168,13 +168,13 @@ func (r *Assignments) Get(e *env.Environment, docs bool, code bool, files []gith
 	return reviewers
 }
 
-func (r *Assignments) getDocsReviewers(author string) []string {
-	setA, setB := getReviewerSets(author, "Core", r.c.DocsReviewers, r.c.DocsReviewersOmit)
+func (r *Assignments) getDocsReviewers(e *env.Environment) []string {
+	setA, setB := getReviewerSets(e, "Core", r.c.DocsReviewers, r.c.DocsReviewersOmit)
```
bernardjkim, line 172: Is the team always supposed to be "Core" here?

gozer: Sort of, the logic for handling docs Pr specifically only really applies to core.

It's something we've purposefully ignored/avoided for the cloud repo

PR #33, opened 2022-10-25
Mark CHANGELOG changes as docs only

ptgott:
Currently, changes to `CHANGELOG.md` are treated as code changes.

PRs to the docs in `gravitational/teleport` often need to edit `CHANGELOG.md` as well, since `docs/pages/changelog.mdx` includes `CHANGELOG.md` as a partial, and it's often necessary to, say, fix links in `CHANGELOG.md` in order to get the docs linter to pass.

If `CHANGELOG.md` is treated as a code change, docs PRs that edit this file need to get two more approvals in order to be merged, even if they don't touch any of the `gravitational/teleport` code.

This change treats `CHANGELOG.md` as a special case for determining that a PR is a docs PR. Other Markdown files outside of `docs/` still require code approval, since the assumption is that these are only relevant to the Engineering team.

### DIFF
--- a/bot/internal/bot/bot.go
+++ b/bot/internal/bot/bot.go
@@ -124,7 +124,8 @@ func classifyChanges(e *env.Environment, files []github.PullRequestFile) (docs b
 	switch e.Repository {
 	case env.TeleportRepo:
 		for _, file := range files {
-			if strings.HasPrefix(file.Name, "docs/") {
+			if strings.HasPrefix(file.Name, "docs/") ||
+				file.Name == "CHANGELOG.md" {
 				docs = true
 			} else {
 				code = true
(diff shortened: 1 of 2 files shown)


bot/internal/bot/bot.go
```diff
@@ -124,7 +124,8 @@ func classifyChanges(e *env.Environment, files []github.PullRequestFile) (docs b
 	switch e.Repository {
 	case env.TeleportRepo:
 		for _, file := range files {
-			if strings.HasPrefix(file.Name, "docs/") {
+			if strings.HasPrefix(file.Name, "docs/") ||
+				file.Name == "CHANGELOG.md" {
```
r0mant, line 128: What if we just check all .md files? E.g. something like
```suggestion
				strings.HasSuffix(file.Name, ".md") {
```

ptgott: I think that depends on if we want to bring RFDs under the purview of the docs team in addition to other Markdown docs in one of the repos that uses the shared workflow bot. The docs team already edits `docs/` files and `CHANGELOG.md` regularly, but we don't tend to work on other Markdown files.

xinding33 (approved):

PR #50, opened 2022-11-30
Delete the remote branch when backport workflow completes

zmb3:
This extends our support for cleaning up remote branches.

The original behavior is that we would only delete the remote branch when backport PRs are merged to one of the release branches.

This means that remote branches aren't cleaned up when:
- PRs get merged to master
- Manual backports get merged to a release branch

These changes ensure that the remote branch is deleted for any internal contributor's PR, whether it is merged to master or a release branch, no matter what the name of the remote branch is.


zmb3:
> Can't we just backport from the `pulls/<n>/head` ref and enable the automatic deletion in github instead? There's no real need to keep the named branch around just for the bot.

I assume we tried that. @russjones got the backports working - do you remember RJ?

PR #51, opened 2022-12-05
bot: label PRs based on their size

zmb3:
This does not take generated files into account.

Fixes gravitational/teleport#14214


bot/internal/bot/bot.go
```diff
@@ -146,7 +159,17 @@ func isLargePR(files []github.PullRequestFile) bool {
 		additions += f.Additions
 		deletions += f.Deletions
 	}
-	return additions-deletions > 1500
+	delta := additions - deletions
+	switch {
+	case delta < 500:
```
r0mant, line 164: Nit: Maybe < 100 for small? It just "feels" like anything over 100 characters is not a trivial changeset.

bot/internal/bot/bot.go
```diff
@@ -137,7 +137,20 @@ func classifyChanges(e *env.Environment, files []github.PullRequestFile) (docs b
...
+	return prSize(files) == xlarge
+}
+
+type sizeLabel string
+
+const (
+	small  sizeLabel = "size-sm"
+	medium sizeLabel = "size-md"
+	large  sizeLabel = "size-lg"
+	xlarge sizeLabel = "size-xl"
```
r0mant, line 150: Nit: What do you think about "namespacing" sizes like we do for other labels like "db/", etc.

```suggestion
	small  sizeLabel = "size/sm"
	medium sizeLabel = "size/md"
	large  sizeLabel = "size/lg"
	xlarge sizeLabel = "size/xl"
```

r0mant (approved):
lgtm with a couple of suggestions.

bot/internal/bot/bot.go
zmb3: Sure!

bot/internal/bot/bot.go
zmb3: Sure, totally open to changing the sizes. (BTW, these are _lines_, not characters)

My goal for this was more to convey to a reviewer when they open a PR how much they have to review. IMO, "small" should mean "I can review this in 10 minutes or less." Do you think 100 is the right number for that?

r0mant: I think it depends on the reviewer tbh. For me personally, 500 lines feels like something I should carve out time for and not an easy fix. 100 feels like something that can be reviewed quickly. We can always experiment and tweak the numbers.

PR #53, opened 2022-12-09
Dependency-Review defaults update

jentfoo:
Update dependency-review to also by default include the `master` branch.  In addition the minimal severity for failing was raised to moderate.

I think this will allow most projects to accept this configuration without modification.

### DIFF
--- a/.github/workflows/dependency-review.yaml
+++ b/.github/workflows/dependency-review.yaml
@@ -28,7 +28,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: 'Dependency Review'
-        uses: actions/dependency-review-action@11310527b429536e263dc6cc47873e608189ba21
+        uses: actions/dependency-review-action@v3
         with:
           fail-on-severity: moderate
           allow-licenses: 0BSD, Apache-2.0, BSD-2-Clause, BSD-2-Clause-FreeBSD, BSD-3-Clause, CC-BY-3.0, CC-BY-4.0, CC0-1.0, ISC, LGPL-2.1, MIT, MPL-2.0, OFL-1.1, Unicode-DFS-2016, Unlicense
--- a/.github/workflows/dependency-review.yaml
+++ b/.github/workflows/dependency-review.yaml
@@ -12,6 +12,7 @@ on:
   pull_request:
     branches:
       - main
+      - master
   workflow_call:
 
 jobs:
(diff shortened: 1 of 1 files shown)


zmb3 (approved):

adaadb6 (approved):

.github/workflows/dependency-review.yaml
```diff
@@ -12,6 +12,7 @@ on:
   pull_request:
     branches:
       - main
+      - master
```
reedloden, line 15: This should be handled on the calling repo, not this repo. Look at how this is called on teleport repo.

.github/workflows/dependency-review.yaml
```diff
@@ -29,4 +30,5 @@ jobs:
       - name: 'Dependency Review'
         uses: actions/dependency-review-action@11310527b429536e263dc6cc47873e608189ba21
         with:
+          fail-on-severity: moderate
```
reedloden, line 33: Do we have cases where we have low severity issues causing problems?

.github/workflows/dependency-review.yaml
```diff
@@ -29,4 +30,5 @@ jobs:
       - name: 'Dependency Review'
         uses: actions/dependency-review-action@11310527b429536e263dc6cc47873e608189ba21
```
reedloden, line 31: If you're updating things, mind changing this?

```suggestion
        uses: actions/dependency-review-action@v3
```

.github/workflows/dependency-review.yaml
jentfoo: Ok, I was trying to see if we could reduce how many repo's needed the branch definition.  If we want it to be defined on the calling repo always should I remove the branches entirely?

.github/workflows/dependency-review.yaml
jentfoo: As best I can tell the teleport repo is the only repo that has this rolled out so far.  So no, but I was concerned it may cause some noise as we roll it out.  If you want to start stricter I am fine with it, I just wondered if it had been considered.

.github/workflows/dependency-review.yaml
jentfoo: Committed.  I wondered why we used the hash here but not anywhere else.

reedloden (approved):

PR #64, opened 2023-01-23
Consult GitHub org membership when identifying internal authors

zmb3:
If a PR author is not explicitly listed as an internal author in the reviewers secret, check whether the author is a member of the GitHub org.

This prevents us from having to maintain (and offboard) a list of every internal author, which is especially useful for employees from sales and CS who occaisonally submit PRs.


wadells (approved):
LGTM.  Thanks for this!

jentfoo (approved):
Looks like a good improvement to me :+1:

PR #65, opened 2023-01-26
Fix an issue that prevents check from succeeding for external authors

zmb3:
The Go GitHub API that we use returns an error for repsonses with an HTTP status code not in the 200-299 range. This caused us to return an error rather than assume the user is an external author.

I noticed this when reviewing the logs of a recent run: https://github.com/gravitational/teleport/actions/runs/4018334113/jobs/6903826941

### DIFF
--- a/bot/internal/github/github.go
+++ b/bot/internal/github/github.go
@@ -502,7 +502,15 @@ func (c *Client) IsOrgMember(ctx context.Context, user string, org string) (bool
 	if err != nil {
 		return false, trace.Wrap(err)
 	}
+
 	resp, err := c.client.Do(ctx, req, nil)
+	// the go-github API returns an error if the request completed
+	// succesfully but returned a non-200 response code, so we attempt
+	// to check the response before checking the error
+	if resp != nil && resp.StatusCode != 0 {
+		return resp.StatusCode == http.StatusNoContent, nil
+	}
+
 	if err != nil {
 		return false, trace.Wrap(err)
 	}


r0mant (approved):

wadells (approved):

bot/internal/github/github.go
```diff
@@ -502,7 +502,15 @@ func (c *Client) IsOrgMember(ctx context.Context, user string, org string) (bool
 	if err != nil {
 		return false, trace.Wrap(err)
 	}
+
 	resp, err := c.client.Do(ctx, req, nil)
+	// the go-github API returns an error if the request completed
+	// succesfully but returned a non-200 response code, so we attempt
+	// to check the response before checking the error
+	if resp != nil && resp.StatusCode != 0 {
```
jentfoo, line 510: This `!= 0` check looks suspicious.  Should this be `!= 200`, or is this zero condition because no response code has been specified?  If the `!= 0` is accurate a comment as to why it would be zero would be helpful.

jentfoo (approved):
Approved, but one question

bot/internal/github/github.go
zmb3: `StatusCode` is of type `int`, and the zero value for an `int` in Go is 0.

The if statement is attempting to make sure there is a legitimate status code to compare to, as opposed to comparing against an "unset" status code. Perhaps a little more defensive than necessary.

jentfoo: I assumed it was a check for the default value, but I have never seen a library which has the response code unset on error responses.  So I just wanted to confirm this was the intention.  Thank you!

zmb3: Yeah, the behavior of this library caught me by surprise too. If there's a non 2xx response, then information is packed into the error return value instead of the response return value. 🤷‍♂️

PR #82, opened 2023-03-03
Automatically add backport labels to docs PRs

ptgott:
Closes gravitational/teleport#15103

We typically need to backport docs PRs to at least one `branch/v[0-9]+` branch unless a docs PR only applies to the next (not current) major Teleport release.

Currently, docs authors need to remember to add backport labels to their PRs so the automatic backport workflow can create branches. If an author forgets to add the correct labels, someone needs to remember to complete the backport some other way. If an author forgets to backport a PR, we need to wait until the next major release before it becomes visible in the default version of the docs site.

This change adjusts the `*Bot.Label` functionality of the workflow bot to label docs PRs with `backport/branch/v[0-9]+`  for the current and past two versions. This way, it's less likely for an author to forget to label their PR with backport labels.

If a branch does not need to be backported, an author can remove the labels. If an author forgets to remove the labels, they can always decline to create a backport branch when we merge the PR, we can delete the backport branch, etc.

- Add a `GitHub.GetLatestReleaseMajorVersion` function that gets the latest Teleport release major version, e.g., `12`
- Add a parameter to `*Bot.labels` for the major version. If this is present, and the PR is a docs PR, the workflow will add `backport` labels for all supported versions.


bot/internal/bot/label.go
```diff
@@ -73,6 +85,23 @@ func (b *Bot) labels(ctx context.Context, files []github.PullRequestFile) ([]str
 			continue
 		}
 
+		docs, _, err := classifyChanges(b.c.Environment, files)
+		if err != nil {
+			return nil, trace.Wrap(err)
+		}
+
+		if docs {
+			labels = append(labels, "documentation")
```
r0mant, line 94: I would keep the original logic that attaches "documentation" label based on prefix and instead of calling "classifyChanges" above again, just looked at attached labels after the "for" loop below and added backport labels if "documentation" label was added by the prefix logic.

bot/internal/bot/label.go
```diff
@@ -100,7 +129,6 @@ func deduplicate(s []string) []string {
 
 var prefixes = map[string][]string{
 	"bpf/":                {"bpf"},
-	"docs/":               {"documentation"},
```
r0mant, line 103: I'd keep this as mentioned above, would make the logic cleaner.

bot/internal/github/github.go
```diff
@@ -644,6 +646,34 @@ type Job struct {
 	ID int64
 }
 
+// extractMajorVersionFromTagName takes a extracts the major version segment of
```
r0mant, line 649:
```suggestion
// extractMajorVersionFromTagName extracts the major version segment of
```

bot/internal/github/github.go
```diff
@@ -644,6 +646,34 @@ type Job struct {
...
+// a Semantic Versioning version string.
+func extractMajorVersionFromTagName(tag string) (int, error) {
+	s := semver.Major(tag)
+	if s == "" {
+		return 0, trace.Wrap(fmt.Errorf("cannot find a major version within tag name %v, which must be formatted using semantic versioning", tag))
```
r0mant, line 654:
```suggestion
		return 0, trace.NotFound("cannot find a major version within tag name %v, which must be formatted using semantic versioning", tag)
```

bot/internal/bot/label.go
```diff
@@ -36,7 +40,14 @@ func (b *Bot) Label(ctx context.Context) error {
 		return trace.Wrap(err)
 	}
 
-	labels, err := b.labels(ctx, files)
+	v, err := b.c.GitHub.GetLatestReleaseMajorVersion(ctx,
+		b.c.Environment.Organization,
+		b.c.Environment.Repository)
+	if err != nil {
+		return trace.Wrap(err)
+	}
```
r0mant, line 48: Can this be done inside `b.labels()` and only on documentation PRs to avoid extra API call for all other PRs? Then you also won't need to change the `b.labels()` method contract.

Also, when you move it, give variable a meaningful name instead of just `v`.

bot/internal/bot/label.go
```diff
@@ -73,6 +85,23 @@ func (b *Bot) labels(ctx context.Context, files []github.PullRequestFile) ([]str
 			continue
 		}
 
+		docs, _, err := classifyChanges(b.c.Environment, files)
+		if err != nil {
+			return nil, trace.Wrap(err)
+		}
+
+		if docs {
```
r0mant, line 93: I think this logic will run on any PR that contains docs changes adding backport labels. Is this intended? Should we run it on PRs that contain _only_ documentation, without any code changes?

r0mant (changes requested):

bot/internal/bot/label.go
zmb3: FWIW, I don't think we should merge this. The point of manually adding labels was to allow the author to specify when a backport is necessary and to which branches.

If we're going to just put the labels on all PRs, why not just try to automatically backport all PRs?

r0mant: @zmb3 I've been having the same reservations to be honest reviewing this. I'm not really sure I understand what problem this is trying to solve, this feels a little backwards tbh. From the description:

>  docs authors need to remember to add backport labels to their PRs so the automatic backport workflow can create branches

This is by design, no? You add backport label to backport your PR. If you forget, you backport manually. 

> If an author forgets to backport a PR, we need to wait until the next major release before it becomes visible in the default version of the docs site.

I'm not sure why do you need to wait? You can just backport and it will get deployed once the PR merges?

@ptgott @alexfortuno

alexfornuto: Just mentioning here for historical purposes, this discussion also happened in Slack and we responded there.

ptgott:
The problem we're trying to solve is ensuring that docs PRs are backported properly, given that Engineering is taking ownership of new feature documentation, and new (or infrequent) docs authors may not be familiar with the backport process. Currently, the Docs Team needs to manually keep track of PRs to ensure that no backports are missing. 

> This is by design, no? You add backport label to backport your PR. If you forget, you backport manually.

We've had a number of cases recently where Alex and I have needed to remind authors to add backport labels, we've added them ourselves, or we have needed to run manual backports. It's not so much the backport process that's the issue, but the need to keep track of backports and remind docs authors. I'm not sure how sustainable this is, given that Alex and I don't own the docs. Ideally, docs processes should run themselves as much as possible without our supervision.

> I'm not sure why do you need to wait? You can just backport and it will get deployed once the PR merges?

If someone forgets to backport, and there's nothing to remind them about it, the backport will only go live once we cut a release from `master` and make that the default docs branch.

At the same time, I understand that leaving unneeded backport labels in docs PRs can be a strain on our CI and create unnecessary branches for us to clean up, so I'm happy to explore other approaches, such as:

- A checklist in a PR template for ensuring the docs PR is complete (including adding backports, running the dev server, etc.)
- A comment that the bot can leave on docs PRs with the checklist (if the PR template isn't effective)
- More training for new docs contributors

PR #86, opened 2023-03-08
Actually run tfsec

jof:
It was discovered that we weren't actually running tfsec in this workflow.

Since we are working to migrate to trivy, there might not be a lot of value in merging this, but this commit will be useful for testing and comparing against what tfsec _would_ have found.

### DIFF
--- a/.github/workflows/terraform-lint.yaml
+++ b/.github/workflows/terraform-lint.yaml
@@ -36,6 +36,11 @@ jobs:
         uses: actions/checkout@v3
 
       - name: tfsec
+        uses: aquasecurity/tfsec-action@v1.0.3
+        with:
+          format: sarif
+
+      - name: tfsec PR Commenter
         uses: aquasecurity/tfsec-pr-commenter-action@v1.3.1
         with:
           github_token: ${{ github.token }}


jentfoo (approved):
Interesting, we do have some tfsec results in github: https://github.com/gravitational/cloud/security/code-scanning?query=is%3Aclosed+branch%3Amaster+tool%3Atfsec

But this change looks reasonable

.github/workflows/terraform-lint.yaml
```diff
@@ -36,6 +36,11 @@ jobs:
         uses: actions/checkout@v3
 
       - name: tfsec
+        uses: aquasecurity/tfsec-action@v1.0.3
```
reedloden, line 39: `aquasecurity/tfsec-pr-commenter-action` runs tfsec itself, so this isn't needed.

See https://github.com/aquasecurity/tfsec-pr-commenter-action/blob/7a44c5dcde5dfab737363e391800629e27b6376b/entrypoint.sh#L66

reedloden (changes requested):

.github/workflows/terraform-lint.yaml
jof: Huh -- I missed that while reviewing the tfsec-pr-commenter-action while trying to figure out why we weren't seeing results.

Still a bit mysterious; we'll work on trivy instead and close this.

PR #87, opened 2023-03-11
Set the `trivy` `scan-type`

jof:
This option is needed to scan the config files the cloned repository files in the workspace (vs scanning images or remote repos).

### DIFF
--- a/.github/workflows/trivy.yaml
+++ b/.github/workflows/trivy.yaml
@@ -20,6 +20,7 @@ jobs:
       - name: Run Trivy vulnerability scanner
         uses: aquasecurity/trivy-action@0.9.1
         with:
+          scan-type: config
           hide-progress: false
           format: sarif
           output: 'trivy-results.sarif'


.github/workflows/trivy.yaml
```diff
@@ -20,6 +20,7 @@ jobs:
       - name: Run Trivy vulnerability scanner
         uses: aquasecurity/trivy-action@0.9.1
         with:
+          scan-type: config
```
reedloden, line 23: Does this need to be `'config'` or not?

.github/workflows/trivy.yaml
```diff
@@ -20,6 +20,7 @@ jobs:
       - name: Run Trivy vulnerability scanner
         uses: aquasecurity/trivy-action@0.9.1
```
reedloden, line 21: As mentioned, I think we should track `master` here.

reedloden (approved):

jentfoo (approved):

.github/workflows/trivy.yaml
jentfoo: To add to @wadells's comment, I would prefer auto-merging Dependabot updates vs tracking the master branch.  But currently hitting the merge button is not a significant overhead.

jof: I'm ambivalent, as this scanner pulls down a vulnerability database every time it runs. It seems to me like this is more about updating the action and test-driving code.

PR #94, opened 2023-03-23
dependency-review: Add option to allow specified GHSAs

justinas:
In https://github.com/gravitational/teleport-plugins/pull/793, due to import of `gravitational/teleport` which does not conform to Go module versioning standard, dependency review is raising a false positive. I am trying to work around this by ignoring this specific GHSA.

See: https://github.com/gravitational/teleport-plugins/pull/793#issuecomment-1481130877 and https://gravitational.slack.com/archives/C0DF0TPMY/p1679580008353179

### DIFF
--- a/.github/workflows/dependency-review.yaml
+++ b/.github/workflows/dependency-review.yaml
@@ -13,6 +13,12 @@ on:
     branches:
       - main
   workflow_call:
+    inputs:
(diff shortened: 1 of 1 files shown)


jentfoo (approved):
Thank you!

PR #115, opened 2023-05-18
Add workflow to exclude flaky tests

zmb3:
Teleport's flaky test detector looks for new or changed tests, and runs them many times in an attempt to detect flakiness. It's not very smart, so sometimes an inconsequential change to an old test (like renaming a variable or changing a function signature) will result in the detector trying to run a slow test too many times and timing out, blocking the team from merging the change.

This workflow will look at a PR and find /excludeflake comments from admins. This is intended to be used as a prerequisite to the flaky test detector, so that we can omit certain changes from flaky test detection.

Note: the flaky test detector itself lives in the Teleport repo. This will run first and pass output to the flaky test detector step.

See corresponding Teleport PR here: https://github.com/gravitational/teleport/pull/26542

### DIFF
--- a/bot/internal/bot/flake.go
+++ b/bot/internal/bot/flake.go
@@ -1,0 +1,56 @@
+package bot
+
+import (
+	"context"
+	"os"
+	"strings"
+
+	"github.com/gravitational/shared-workflows/bot/internal/github"
+	"github.com/gravitational/trace"
+)
+
+const skipPrefix = "/excludeflake"
+
+// ExcludeFlakes gets the list of test names that can be
+// excluded from flaky test detection for a particular PR.
+// Admin reviewers can exclude tests by commenting on a
+// PR with "/excludeflake Test1 Test2".
+//
+// The result is written to a GitHub Action output parameter
+// named FLAKE_SKIP.
+func (b *Bot) ExcludeFlakes(ctx context.Context) error {
+	skip, err := b.testsToSkip(ctx)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	outfile := os.Getenv(github.OutputEnv)
+	err = os.WriteFile(outfile, []byte("FLAKE_SKIP="+strings.Join(skip, " ")), 0644)
+	return trace.Wrap(err)
+}
+
+func (b *Bot) testsToSkip(ctx context.Context) ([]string, error) {
+	comments, err := b.c.GitHub.ListComments(ctx,
+		b.c.Environment.Organization,
+		b.c.Environment.Repository,
+		b.c.Environment.Number)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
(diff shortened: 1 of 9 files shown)


bot/internal/bot/flake.go
```diff
@@ -0,0 +1,56 @@
...
+// The result is written to a GitHub Action output parameter
+// named FLAKE_SKIP.
+func (b *Bot) ExcludeFlakes(ctx context.Context) error {
+	skip, err := b.testsToSkip(ctx)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	outfile := os.Getenv(github.OutputEnv)
+	err = os.WriteFile(outfile, []byte("FLAKE_SKIP="+strings.Join(skip, " ")), 0644)
```
zmb3, line 28: I'm not sure if a space is valid in GHA outputs, but it's the only character that I could come up with that won't occur in a Go test name (even subtest descriptions replace ` ` with `_`).

r0mant: How about tilde `~`? Space is fine too if it works.

bot/internal/github/github.go
```diff
@@ -561,8 +565,15 @@ func (c *Client) CreateComment(ctx context.Context, organization string, reposit
 	return nil
 }
 
-// ListComments lists all comemnts on an issue or PR.
-func (c *Client) ListComments(ctx context.Context, organization string, repository string, number int) ([]string, error) {
+// Comment represents an "issue comment" on a GitHub issue or pull request.
+// This does not include comments that are part of reviews.
+type Comment struct {
+	Author string // the GitHub username of the author
+	Body   string // the text of the comment
```
r0mant, line 572: Nit: Use godoc-style comments instead on top of the fields?

r0mant (approved):

r0mant (approved):

bot/internal/github/github.go
zmb3: godoc still renders it nicdely, and I kind of like when the `string` is aligned.

```
➜ go doc "github.com/gravitational/shared-workflows/bot/internal/github.Comment"
package github // import "github.com/gravitational/shared-workflows/bot/internal/github"

type Comment struct {
	Author string // the GitHub username of the author
	Body   string // the text of the comment
}
    Comment represents an "issue comment" on a GitHub issue or pull request.
    This does not include comments that are part of reviews.


shared-workflows/bot on zmb3/get-flake-excludes
➜ go doc "github.com/gravitational/shared-workflows/bot/internal/github.Comment.Author"
package github // import "github.com/gravitational/shared-workflows/bot/internal/github"

type Comment struct {
    Author string  // the GitHub username of the author

    // ... other fields elided ...
}
```

bot/internal/bot/flake.go
zmb3: Good idea, will do.

PR #144, opened 2023-07-19
Add BlueOak-1.0.0

benarent:
New license for https://github.com/gravitational/docs/actions/runs/5581482811/jobs/10199681165 

More info for the BlueOak License https://blueoakcouncil.org/license/1.0.0

### DIFF
--- a/.github/workflows/dependency-review.yaml
+++ b/.github/workflows/dependency-review.yaml
@@ -36,5 +36,5 @@ jobs:
         uses: actions/dependency-review-action@v3
         with:
           fail-on-severity: moderate
-          allow-licenses: 0BSD, Apache-2.0, BSD-2-Clause, BSD-2-Clause-FreeBSD, BSD-3-Clause, CC-BY-3.0, CC-BY-4.0, CC0-1.0, ISC, LGPL-2.1, MIT, MIT-0, MPL-2.0, ODC-By-1.0, OFL-1.1, Python-2.0, Unicode-DFS-2016, Unlicense, Zlib, (MIT OR Apache-2.0) AND Unicode-DFS-2016, Apache-2.0 AND BSD-3-Clause, ISC AND MIT, MIT AND Zlib, MIT AND BSD-3-Clause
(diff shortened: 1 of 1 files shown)


jentfoo:
The FAQ on the license does not inspire confidence with regards to compatibility of Blue Oak: https://blueoakcouncil.org/license-faq

They also require their license to be attributed (granted we could probably improve attribution generally):
> You must ensure that everyone who gets a copy of any part of this software from you, with or without changes, also gets the text of this license or a link to https://blueoakcouncil.org/license/1.0.0.

I also was not able to find if OSI or other similar organizations have reviewed this license.  I assume it has not been reviewed yet by them.

That said, it's otherwise an attribution only license.  But there is enough nuance that I wonder if this should be reviewed by legal.

@benarent Do you think we could exclude the `jackspeak` library safely?  This appears to do argument parsing, so I wondered if it's used or could be replaced.

wadells:
Reading through it, this license seems reasonable, permissive, and readable(!) IMO.  It looks like a [BSD 2 Clause](https://spdx.org/licenses/BSD-2-Clause-Patent.html), but more readable.

> They also require their license to be attributed (granted we could probably improve attribution generally):

A legitimate problem we have, but not speicific to blueoak.
 
> I was not able to find if OSI or other similar organizations have reviewed this license. I assume it has not been reviewed yet by them.

It doesn't look reviewed as of 2023-07.  The authors appear to be boycotting or otherwise fed up with the process:

"I have no current plans to submit the license primarily because I am too busy to have a massively inefficient discussion on license-review. But I'd be willing to make that time commitment if I thought it was helping prototype something new and better." - Luis

https://lists.opensource.org/pipermail/license-discuss_lists.opensource.org/2019-March/020279.html

Luis is one of the directors on the board: https://blueoakcouncil.org/about.

> That said, it's otherwise an attribution only license. But there is enough nuance that I wonder if this should be reviewed by legal.

Interestingly, the blueoak project seems to be trying to formalize what might be approved by legal. Take a look at this page:

https://blueoakcouncil.org/company-policy

IANAL, but if I were, this looks pretty promising as a list of approved licenses.

wadells (commented):
Leaving a comment to drop this out of my review queue -- I forgot to do my earlier feedback via the "Review changes" button.

reedloden (approved):
Spoke to @TMG63 about this, and fine for the case where it's being used.

PR #148, opened 2023-08-03
Create backport PR automatically

r0mant:
2 changes to the backport bot:

- Attempt to create the backport PR automatically. Fallback to the current "quick PR" link if it fails.
- Transfer all labels (except for "backport/" ones) to the opened backport PR.

### DIFF
--- a/bot/internal/bot/backport.go
+++ b/bot/internal/bot/backport.go
@@ -97,29 +97,51 @@ func (b *Bot) Backport(ctx context.Context) error {
 		if err != nil {
 			log.Printf("Failed to create backport branch:\n%v\n", trace.DebugReport(err))
 			rows = append(rows, row{
-				Branch: base,
-				Failed: true,
-				Link:   u,
+				Branch:       base,
+				BranchFailed: true,
+				Link:         u,
 			})
 			continue
 		}
 
+		backportPR, err := b.createBackportPR(ctx, pull, head, base)
+		if err != nil {
+			// If we failed to create a backport PR automatically, fallback to
+			// giving a "quick PR" link since the backport branch was created
+			// successfully.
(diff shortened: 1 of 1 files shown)


jentfoo (approved):

r0mant:
@jentfoo @reedloden A side effect of this change would be that the original PR creator could technically approve their own backport PR because it will be opened by the bot account. We would still require the 2nd approval.

Just want to make sure it's fine from the compliance perspective like SOC2, etc.

jentfoo:
@r0mant, brainstorming, what if we protected the automatic backport branches so that custom changes can't be pushed to them?  If that's reasonable then I don't think there is any concerns, since we have control over what content can be in the PR.

r0mant:
@jentfoo Yeah, we can create a branch protection rule that will require a pull request to merge to `backport/xxx` branches. I'll send a PR to github-terraform.

r0mant:
@jentfoo Before I do it though: do we _have to do it_? I.e. my question was, will we actually violate any compliance provisions if we don't protect backport branches? Just trying to not introduce any extra barriers for the team unless really necessary.

reedloden:
From a purely compliance standpoint, as long as we have at least one peer review (as in, review from an authorized non-bot approver who is not the code author), we're fine. The two separate reviews is a goal, but not a hard requirement.

From a security standpoint, adding branch protections is the right call. Lack of branch protections has bitten us in the past, and I'm concerned what would happen if a non-bot user pushed branches that matched the same pattern. How does the bot handle it if a branch already exists? What if somebody changes the underlying branch contents (at any time during the process)?

r0mant:
@reedloden If non-bot user pushes `backport/` branch, nothing would change compared to now. Since it's a user that pushes it, it will require 2 approvals from others to merge to master/release branches just like now.

PR #150, opened 2023-08-07
Add binary bloat checking

rosstimothy:
A new `bloat` target is added to the bot to determine if binaries before and after a change cause an increase in size. All changes less than +1MB are ignored, a warning is emitted for any changes >=+1MB, and >=+3MB results in an error.

An increase of >=3MB is allowed if an admin leaves a comment on the PR with the prefix `/excludebloat` similar to how flaky tests can be skipped.

### DIFF
--- a/bot/internal/bot/bloat.go
+++ b/bot/internal/bot/bloat.go
@@ -1,0 +1,177 @@
+package bot
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"unicode/utf8"
+
+	"github.com/gravitational/trace"
+)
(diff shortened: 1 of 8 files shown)


bot/internal/bot/bloat.go
```diff
@@ -0,0 +1,177 @@
...
+	"strings"
+	"unicode/utf8"
+
+	"github.com/gravitational/trace"
+)
+
+const (
+	// skipArtifactPrefix the comment prefix to use in order to
+	// skip particular artifacts from the bloat check.
+	skipArtifactPrefix = "/excludeartifact"
```
zmb3, line 20:
```suggestion
	skipBloatCheckPrefix = "/excludebloat"
```
?

PR #157, opened 2023-08-29
RFD 0001: Tooling Requirements

fheinecke:
Rendered: https://github.com/gravitational/shared-workflows/blob/fred/rfd-0001/rfd/0001-tooling-requirements.md


rfd/0001-tooling-requirements.md
```diff
@@ -0,0 +1,150 @@
+---
+authors: Fred Heinecke (fred.heinecke@goteleport.com)
+state: draft
+---
+
+# RFD 1 - Tooling Requirements
+
+## Required approvers
+
+* Engineering: (@r0mant)
```
fheinecke, line 10: I am not quite sure who should be listed here. Does any reviewers have suggestions?

rfd/0001-tooling-requirements.md
```diff
@@ -0,0 +1,150 @@
...
+1. PRs are filed to add/modify the project.
+2. The project's CI workflow triggers to validate the changes, codeowners approve, and PRs are merged.
+3. Artifacts are built for each PR to facilitate release testing.
+4. A new PR is filed and merged that updates the changelog to a new version.
+5. The release workflow is triggered, which creates a new tag and publishes a fresh build.
+
+### Dependency management
+Historically this repo has used Dependabot to manage dependency updates. Unfortunately however, Dependabot does not support two key "languages" used by Github Actions: actions themselves (used when building composite actions), and Dockerfiles (used when building Dockerfile actions). As a result, [Renovate](https://docs.renovatebot.com/) will be used for new and migrated projects instead. Renovate is highly flexible and can be configured on a per-project basis.
+
+Renovate will self-hosted and configured to run every 12 hours. Updates for each project should be grouped as much as reasonably possible to reduce the number of PRs opened at any given time. Update PRs will be assigned to code owners for each project. Digest pinning will be used wherever supported.
```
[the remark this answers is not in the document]
fheinecke: Oh really? I've had the opposite problem (dependabot not working as needed where renovate does) on the cloud-terraform repo (which now uses self-hosted Renovate) , and just a few minutes ago on the [cloud repo](https://github.com/gravitational/cloud/pull/5797).

What issues have there been with Renovate on `gravitational/teleport`?

rfd/0001-tooling-requirements.md
wadells: You can add `* Security: (@reedloden  || @wadells)`.  I don't know that product will have an opinion here.  Otherwise, I think you've added the right reviwers already.

rfd/0001-tooling-requirements.md
wadells, line 143:
```suggestion
Renovate will self-hosted and configured to run daily. Updates for each project should be grouped as much as reasonably possible to reduce the number of PRs opened at any given time. Update PRs will be assigned to code owners for each project. Digest pinning will be used wherever supported.
```

I don't think we'd realistically merge more than once a day anyhow.  In `teleport`/`teleport.e` we do updates weekly to cut down on sustaining -- and I'm starting to lean that way for `cloud-terraform`.  In my opinion: the time lag for picking up a critical security update is lower risk to Gravitational than the cost of having maintainers spend more time on sustaining dependency updates and less time on new development.

rfd/0001-tooling-requirements.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Security considerations
+The primary source of security concern with anything living in this repo is that the repo is public. The impact of making a mistake with one of our workflows (such as exposing a sensitive value) is much greater than if the repo was private/internal.
+
+I took a look at the forks, stars, watchers, issues, and PRs for this repo. As far as I can tell, all but two or three of the interactions (stars and watchers) from non-Gravitational Github accounts are from bots. Based on this I think it is reasonable to assume that nobody outside of Gravitational uses this repo. Additionally, I don't personally believe that the few existing tools in this repo have a large value to the open source community. With this information in mind, I believe that we make this repo private. Doing so would alleviate most of the security concerns associated with this RFD.
```
wadells, line 148: I believe `shared-workflows` was originally private, and was made public to work with `teleport` and other repos.  See https://github.com/gravitational/github-terraform/pull/240

According to the docs, I don't think we _can_ make this private:

> Actions and reusable workflows stored in private repositories cannot be used in public repositories.

https://docs.github.com/en/actions/creating-actions/sharing-actions-and-workflows-from-your-private-repository#about-github-actions-access-to-private-repositories

As such, we may consider a different (new?) repo.  I don't think `teleport.e` is the best fit for these tools (getting bogged down in teleport CI sucks for orthogonal changes), but nor do we have another appropriate place.  Three years ago, the right place would have been https://github.com/gravitational/ops, but that repo is a skeleton now.  We could resurrect ops -- but I don't see much point as none of the previous history is relevant to your proposed path forward.

wadells (commented):
Thanks for putting this together!

I've added a first round of quick thoughts.  I need to mull over the differences/commonalities between internal golang tooling and our common github actions before I can comment on the structure and requirements.

rfd/0001-tooling-requirements.md
fheinecke: > Actions and reusable workflows stored in private repositories cannot be used in public repositories.

The workaround that I've found for this is to use a workflow dispatch event to call a workflow in another repository... although that is definitely messy and probably not the right approach here.

> As such, we may consider a different (new?) repo

I thought about that, but did not originally suggest it because I figured it'd be more work for you and your team. I am not opposed to that if you think it's the right path forward. All that being said, if actions cannot be used from public repos then switching to a private repo (of any form) might make it harder to adopt this change, or may result in a separate private repo down the line. What are your thoughts here?

rfd/0001-tooling-requirements.md
```diff
@@ -0,0 +1,151 @@
...
+### Project releases
+The process for releasing a new version of a project should be as follows:
+1. PRs are filed to add/modify the project.
+2. The project's CI workflow triggers to validate the changes, codeowners approve, and PRs are merged.
+3. Artifacts are built for each PR to facilitate release testing.
+4. A new PR is filed and merged that updates the changelog to a new version.
+5. The release workflow is triggered, which creates a new tag and publishes a fresh build.
+
+### Dependency management
+Historically this repo has used Dependabot to manage dependency updates. Unfortunately however, Dependabot does not support two key "languages" used by Github Actions: actions themselves (used when building composite actions), and Dockerfiles (used when building Dockerfile actions). As a result, [Renovate](https://docs.renovatebot.com/) will be used for new and migrated projects instead. Renovate is highly flexible and can be configured on a per-project basis.
```
zmb3, line 142: What support for Dockerfiles does renovate offer? Does it update your base image or something?

fheinecke: > Does it update your base image or something?

Yep, exactly that.

rfd/0001-tooling-requirements.md
```diff
@@ -0,0 +1,151 @@
...
+
+#### CD workflow
+This workflow should run when a PR is merged, and optionally upon PR commits as a way to produce test builds. When ran, this workflow should upload artifacts as Github Packages where possible. These artifacts should not be uploaded to AWS unless absolutely necessary, and with Security's permission, as outlined in the [Security considerations section](#security-considerations). The name of these artifacts should include a dev tag generated by `git describe --tags` to ensure uniqueness.
+
+#### Release workflow
+The release workflow should be triggered when a new version is added to the CHANGELOG on `main` branch. When this occurs, the workflow should:
+1. Extract the new version from the CHANGELOG.
```
r0mant, line 128: Seems a little strange to trigger release on changelog update, why not trigger it on a tag?

rfd/0001-tooling-requirements.md
r0mant: Can we just avoid pushing secrets to this repo?

r0mant (approved):
Overall proposal lgtm with a couple of questions, I think @camscale and @tcsc need to review this as well.

rfd/0001-tooling-requirements.md
```diff
@@ -0,0 +1,151 @@
...
+* While Dependabot has historically been used for keeping this repo's tool's dependencies up to date, Renovate will now be added as well. See the [Dependency management](#dependency-management) section for details on why this choice was made. The structure of the configuration will be similar to `gravitational/cloud-terraform`, with a top-level config and individual configs for each tool. This allows Renovate to be specifically configured for each tool.
+* There will be a pull request template for new projects that includes a checklist of all the items listed in [Project requirements](#project-requirements).
+* Some additional boilerplate files will be added such as a `LICENSE` and `SECURITY.md` file. The contents of these files will be copy/pasted from `gravitational/teleport` and tweaked as appropriate.
+
+### Project requirements
+New "projects" (tools and libraries) should live in dedicated directories under `tools` and `libs`, wherever is most appropriate. Each project directory should be treated as it's own separate repo in that all resources associated with the project's lifecycle should be self-contained in the directory.
```
tcsc, line 109: I am anticipating that there will be some `tool`s that depend on `libs`. Are we expecting to have one top-level `go.mod` for both? Or will that be managed by revovate?

...or am I completely barking up the wrong tree?

tcsc (commented):
Looks good to me. I would like to see section/example showing how the tooling in this is going to be _consumed_ by other projects.

rfd/0001-tooling-requirements.md
fheinecke: Good catch - I should have included that. I'd like each tool and library to have their own go.mod, with renovate managing them. Most of the time the dependencies should be the same, but with this structure we should be able to move dependencies forward for most projects even if one project has to be kept behind.

I don't think this will generate an abnormally large number of dependency update PRs as long as we configure Renovate properly to group dependency updates for each project into a single PR.

tcsc: Yeah - seems fair. It would definitely be good to avoid the trap of requiring one module spec for the whole lot and painting ourselves into a corner.

rfd/0001-tooling-requirements.md
fheinecke: I don't think this is strictly needed. Updated to trigger on tag.

rfd/0001-tooling-requirements.md
fheinecke: Updated to show go.mod/go.sum in the example file tree.

PR #159, opened 2023-09-01
Added bot command to enforce changelog requirements

fheinecke:
This allows for repos using the shared-workflows bot to enforce changelog requirements.

### DIFF
--- a/bot/internal/bot/changelog.go
+++ b/bot/internal/bot/changelog.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 Gravitational, Inc.
+Copyright 2023 Gravitational, Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/bot/internal/bot/changelog.go
+++ b/bot/internal/bot/changelog.go
@@ -87,11 +87,11 @@ func (b *Bot) getChangelogEntries(ctx context.Context, prBody string) ([]string,
 func (b *Bot) validateChangelogEntry(ctx context.Context, changelogEntry string) error {
 	changelogEntry = strings.ToLower(strings.TrimSpace(changelogEntry)) // Format the entry for easy validation
 	if changelogEntry == "" {
-		return b.logFailedCheck(ctx, "The changelog entry must contain one or more non-whitespace characters")
+		return b.logFailedCheck(ctx, "The changelog entry must contain one or more non-whitespace characters.")
 	}
 
 	if !unicode.IsLetter([]rune(changelogEntry)[0]) {
-		return b.logFailedCheck(ctx, "The changelog entry must start with a letter")
+		return b.logFailedCheck(ctx, "The changelog entry must start with a letter.")
 	}
 
 	if strings.HasPrefix(changelogEntry, "backport of") ||
@@ -100,15 +100,15 @@ func (b *Bot) validateChangelogEntry(ctx context.Context, changelogEntry string)
 	}
 
 	if strings.Contains(changelogEntry, "](") {
-		return b.logFailedCheck(ctx, "The changelog entry must not contain a Markdown link or image")
+		return b.logFailedCheck(ctx, "The changelog entry must not contain a Markdown link or image.")
 	}
 
 	if strings.Contains(changelogEntry, "```") {
-		return b.logFailedCheck(ctx, "The changelog entry must not contain a multiline code block")
+		return b.logFailedCheck(ctx, "The changelog entry must not contain a multiline code block.")
 	}
 
 	if changelogEntry == "none" {
-		return b.logFailedCheck(ctx, "The %q label must be set instead of listing 'none' as the changelog entry", NoChangelogLabel)
+		return b.logFailedCheck(ctx, "The %q label must be set instead of listing 'none' as the changelog entry.", NoChangelogLabel)
 	}
 
 	return nil
--- a/bot/internal/bot/changelog.go
+++ b/bot/internal/bot/changelog.go
@@ -119,7 +119,7 @@ func (b *Bot) logFailedCheck(ctx context.Context, format string, args ...interfa
 		b.c.Environment.Organization,
 		b.c.Environment.Repository,
 		b.c.Environment.Number,
-		fmt.Sprintf(format, args...),
+		fmt.Sprintf("The PR changelog entry failed validation: %s", fmt.Sprintf(format, args...)),
 	)
 	if err != nil {
 		return trace.Wrap(err, "failed to create or update the changelog comment")
--- a/bot/internal/bot/changelog.go
+++ b/bot/internal/bot/changelog.go
@@ -51,34 +51,35 @@ func (b *Bot) CheckChangelog(ctx context.Context) error {
 		return nil
 	}
 
-	changelogEntry, err := b.getChangelogEntry(ctx, pull.UnsafeBody)
+	changelogEntries, err := b.getChangelogEntries(ctx, pull.UnsafeBody)
 	if err != nil {
 		return trace.Wrap(err, "failed to get changelog entry")
 	}
 
-	err = b.validateChangelogEntry(ctx, changelogEntry)
-	if err != nil {
-		return trace.Wrap(err, "failed to validate changelog entry")
+	for _, changelogEntry := range changelogEntries {
+		err = b.validateChangelogEntry(ctx, changelogEntry)
+		if err != nil {
+			return trace.Wrap(err, "failed to validate changelog entry %q", changelogEntry)
+		}
 	}
 
 	return nil
 }
 
-func (b *Bot) getChangelogEntry(ctx context.Context, prBody string) (string, error) {
+func (b *Bot) getChangelogEntries(ctx context.Context, prBody string) ([]string, error) {
 	changelogRegex := regexp.MustCompile(ChangelogRegex)
 
-	changelogMatches := changelogRegex.FindAllString(prBody, 2) // Two or more changelog entries is invalid, so no need to search for more than two.
+	changelogMatches := changelogRegex.FindAllString(prBody, -1)
 	if len(changelogMatches) == 0 {
-		return "", b.logFailedCheck(ctx, "Changelog entry not found in the PR body. Please add a %q label to the PR, or a changelog line starting with `%s` followed by the changelog entry for the PR.", NoChangelogLabel, ChangelogPrefix)
+		return nil, b.logFailedCheck(ctx, "Changelog entry not found in the PR body. Please add a %q label to the PR, or changelog lines starting with `%s` followed by the changelog entries for the PR.", NoChangelogLabel, ChangelogPrefix)
 	}
 
-	if len(changelogMatches) > 1 {
-		return "", b.logFailedCheck(ctx, "Found two or more changelog entries in the PR body: %v", changelogMatches)
+	for i, changelogMatch := range changelogMatches {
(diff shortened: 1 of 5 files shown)


bot/internal/bot/changelog.go
```diff
@@ -0,0 +1,159 @@
...
+
+	"github.com/gravitational/trace"
+	"k8s.io/utils/strings/slices"
+)
+
+const NoChangelogLabel string = "no-changelog"
+const ChangelogPrefix string = "changelog: "
+const ChangelogTag string = "changelog"
+
+func (b *Bot) Changelog(ctx context.Context) error {
```
r0mant, line 35: Nit: `Changelog` is a very vague name for this method. Maybe `CheckChangelog` or something? Also, add godoc.

bot/internal/bot/changelog.go
```diff
@@ -0,0 +1,159 @@
...
+		b.c.Environment.Organization,
+		b.c.Environment.Repository,
+		b.c.Environment.Number,
+	)
+	if err != nil {
+		return trace.Wrap(err, "failed to retrieve pull request for https://github.com/%s/%s/pull/%d", b.c.Environment.Organization, b.c.Environment.Repository, b.c.Environment.Number)
+	}
+
+	if slices.Contains(pull.UnsafeLabels, NoChangelogLabel) {
+		log.Printf("PR contains no changelog label %q, skipping changelog check", NoChangelogLabel)
```
r0mant, line 46:
```suggestion
		log.Printf("PR contains %q label, skipping changelog check", NoChangelogLabel)
```

bot/internal/bot/changelog.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return trace.Wrap(err, "failed to delete changelog comments")
+	}
+
+	return nil
+}
+
+func (b *Bot) getChangelogEntry(ctx context.Context, prBody string) (string, error) {
+	changelogRegexString := fmt.Sprintf("(?mi)^%s.*", regexp.QuoteMeta(ChangelogPrefix))
+	changelogRegex, err := regexp.Compile(changelogRegexString)
```
r0mant, line 70: I would declare changelog regex as global variable and compile it using regexp.MustCompile. Doesn't really matter from performance perspective since this runs once per job, but would make this function a little cleaner.

bot/internal/bot/changelog.go
```diff
@@ -0,0 +1,159 @@
...
+	if len(changelogMatches) == 0 {
+		return "", b.logFailedCheck(ctx, "Changelog entry not found in the PR body. Please add a %q label to the PR, or a changelog line starting with `%s` followed by the changelog entry for the PR.", NoChangelogLabel, ChangelogPrefix)
+	}
+
+	if len(changelogMatches) > 1 {
+		return "", b.logFailedCheck(ctx, "Found two or more changelog entries in the PR body: %v", changelogMatches)
+	}
+
+	changelogEntry := changelogMatches[0][len(ChangelogPrefix):] // Case insensitive prefix removal
+	log.Printf("Found changelog entry %q matching %q", changelogEntry, changelogRegexString)
```
r0mant, line 85: Nit: Don't think we really need to log the regex.
```suggestion
	log.Printf("Found changelog entry: %q", changelogEntry)
```

bot/internal/bot/changelog.go
```diff
@@ -0,0 +1,159 @@
...
+		return b.logFailedCheck(ctx, "The changelog entry must contain one or more non-whitespace characters")
+	}
+
+	if !unicode.IsLetter([]rune(changelogEntry)[0]) {
+		return b.logFailedCheck(ctx, "The changelog entry must start with a letter")
+	}
+
+	if unicode.IsLower([]rune(changelogEntry)[0]) {
+		return b.logFailedCheck(ctx, "The changelog entry must start with an uppercase character")
+	}
```
r0mant, line 102: This seems excessive tbh, this we can easily fix with other tooling.
```suggestion
```

bot/internal/bot/changelog.go
```diff
@@ -0,0 +1,159 @@
...
+
+	if strings.TrimSpace(changelogEntry) != changelogEntry {
+		return b.logFailedCheck(ctx, "The changelog entry must not contain leading or trailing whitespace")
+	}
```
r0mant, line 106: Would remove this also.
```suggestion
```

bot/internal/bot/changelog.go
```diff
@@ -0,0 +1,159 @@
...
+
+	if strings.HasPrefix(strings.ToLower(changelogEntry), "backport of") ||
+		strings.HasPrefix(strings.ToLower(changelogEntry), "backports") {
+		return b.logFailedCheck(ctx, "The changelog entry must contain the actual change, not a reference to the source PR of the backport.")
+	}
+
+	if unicode.IsPunct([]rune(changelogEntry)[len(changelogEntry)-1:][0]) {
+		return b.logFailedCheck(ctx, "The changelog entry must not end with punctuation.")
+	}
```
r0mant, line 115: And this.
```suggestion
```

bot/internal/bot/bot.go
```diff
@@ -62,6 +62,17 @@ type Client interface {
...
+	// ListStatefulComments finds comments on a given PR matching the provided tag
+	ListStatefulComments(ctx context.Context, organization string, repository string, number int, tag string) ([]github.Comment, error)
+
+	// CreateOrUpdateStatefulComment will create a comment on a Pull Request, or update a pre-existing comment if one is found with a matching tag.
+	// Returns the number of created or updated comments.
+	CreateOrUpdateStatefulComment(ctx context.Context, organization string, repository string, number int, comment string, tag string) (int, error)
+
+	// DeleteStatefulComment will delete all comments on a Pull Request found with a matching tag.
+	// Returns the number of deleted comments.
+	DeleteStatefulComment(ctx context.Context, organization string, repository string, number int, tag string) (int, error)
```
r0mant, line 74: "Stateful" is a little confusing. Maybe "ListCommentsByTag", "DeleteCommentByTag", etc. instead?

bot/internal/bot/changelog.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return trace.Wrap(err, "failed to get changelog entry")
+	}
+
+	err = b.validateChangelogEntry(ctx, changelogEntry)
+	if err != nil {
+		return trace.Wrap(err, "failed to validate changelog entry")
+	}
+
+	err = b.deleteComments(ctx)
```
r0mant, line 60: TBH I would consider simplifying significantly, just check the PR description and leave a comment if changelog is missing or incorrect, without trying to update or delete existing comments. Wouldn't need to introduce the concept of "stateful" comments either then.

r0mant (commented):
@fheinecke Overall I like this, but left a few simplification suggestions.

bot/internal/bot/changelog.go
fheinecke: I can do this but then I have to hard code the entire regex string (`(?mi)^changelog: .*` in this case) as I cannot call functions when defining a const. The downside of this is that it would be _slightly_ harder to update the changelog regex in the future, and makes the pattern matching logic a little harder to read (IMO). Would you like me to continue moving the regex string to a const?

bot/internal/bot/changelog.go
fheinecke: I somewhat disagree here - I have found regex logging to be helpful in the past, especially when testing an update to regex logic. That being said, if the regex is hard-coded as discussed above then you are right - this log entry is probably unnecessary.

bot/internal/bot/changelog.go
fheinecke: I can do this, but in some cases it will result in PRs having several bot comments that may make the PR webpages harder to read. Any time a label or the PR is added/removed/changed a new comment will be added. This will be especially impactful when we initially implement the changelog check. Is the code simplification worth the increase in comments?

bot/internal/bot/changelog.go
```diff
@@ -0,0 +1,123 @@
+/*
+Copyright 2022 Gravitational, Inc.
```
r0mant, line 2:
```suggestion
Copyright 2023 Gravitational, Inc.
```

bot/internal/bot/changelog.go
```diff
@@ -0,0 +1,123 @@
...
+		return trace.Wrap(err, "failed to validate changelog entry")
+	}
+
+	return nil
+}
+
+func (b *Bot) getChangelogEntry(ctx context.Context, prBody string) (string, error) {
+	changelogRegex := regexp.MustCompile(ChangelogRegex)
+
+	changelogMatches := changelogRegex.FindAllString(prBody, 2) // Two or more changelog entries is invalid, so no need to search for more than two.
```
r0mant, line 70: Nit: TBH I think we could simplify this further and just take the first one in case multiple are found (which would probably be rare if will ever happen at all) but that's up to you.

bot/internal/bot/changelog.go
```diff
@@ -0,0 +1,123 @@
...
+	changelogEntry := changelogMatches[0][len(ChangelogPrefix):] // Case insensitive prefix removal
+	log.Printf("Found changelog entry %q", changelogEntry)
+	return changelogEntry, nil
+}
+
+// Checks for common issues with the changelog entry.
+// This is not intended to be comprehensive, rather, it is intended to cover the majority of problems.
+func (b *Bot) validateChangelogEntry(ctx context.Context, changelogEntry string) error {
+	if strings.TrimSpace(changelogEntry) == "" {
+		return b.logFailedCheck(ctx, "The changelog entry must contain one or more non-whitespace characters")
```
r0mant, line 88: To make logic more clear, I would make the following changes:

- Instead of returning `b.logFailedCheck()` from each branch here and in "getChangelogEntry", have them return a normal `trace.BadParameter("xxx")` error.
- In `CheckChangelog` call `b.logFailedCheck(ctx, err)` when the respective function returns an error.

Basically, move `logFailedCheck` call up the stack. Otherwise it's very implicit that get/validate functions also leave comments on the PR.

bot/internal/bot/changelog.go
fheinecke: I'm going to remove this altogether. It's pretty rare, but @camscale ran into an issue last week (maybe two weeks ago) on a backport PR where multiple changelog entries were listed. This can sometimes occur when a backport contains multiple master branch PRs.

It should be trivial for downstream tooling to handle multiple changelog entries.

bot/internal/bot/changelog.go
fheinecke: I disagree here, on the grounds that `validateChangelogEntry` could potentially return an error that is not due to an invalid changelog. Currently this is not an issue, however, if a developer were to add another check that does something like checking to see if the changelog entry is already in the CHANGELOG.md file, then this could result in an error message being written to the PR body that is not due to an invalid changelog entry. More specifically, if a check was added that did something like `err := os.ReadFile("CHANGELOG.md")` then this error could potentially get logged as a changelog entry failure.

Maybe one of these would be a better solution to the issue:
* I could add a comment to the `validateChangelogEntry` function that explicitly says that it logs errors to the PR
* I could return an error type that is specific to a changelog entry problem, and only log to the PR if the error is of this type

What do you think?

If I were to go with the second route, would it be better to return and instance of this new error type, or return a `trace.Wrap` of this new error type? In other words, `ValidationError("some message")` or `trace.Wrap(ValidationError("some message"), "some error occured")`?

r0mant (approved):

PR #172, opened 2023-10-20
Update go modules to address dependency CVE

fheinecke:
The new Orca tool is reporting a 7.5 CVE in the Golang yaml dependency. Most of the dependencies are a little out of date so I've gone ahead and bumped them all.

We don't have any tests for the bot, but it still compiles ¯\\_(ツ)_/¯


jentfoo (approved):

reedloden (approved):

jentfoo:
We should add this project to the Dependabot config as well: https://github.com/gravitational/shared-workflows/blob/main/.github/dependabot.yml

reedloden:
> We should add this project to the Dependabot config as well: https://github.com/gravitational/shared-workflows/blob/main/.github/dependabot.yml

That's for automated version updates, no? I thought security updates were separate.

jentfoo:
@reedloden Sorta.  If you have no Dependabot configuration at all you get CVE only updates.  But I believe once you create a Dependabot configuration it will only scan projects which are defined in that config.  If we still only want CVE updates we need to set the `open-pull-requests-limit	` to `0`

PR #179, opened 2023-11-10
Allow preferred docs reviewers

ptgott:
The code reviewer configuration allows us to name preferred reviewers for certain file paths. For docs PRs, the workflow bot does not consult the preferred reviewer configuration. This means that, even if we add the `preferredReviewerFor` key to a `codeReviewers` entry with a docs path as a value, the workflow bot continues to add only `docsReviewers` to docs PRs.

In this change, the workflow bot gets docs reviewers by:

1. Retrieving the docs reviewer pool from the config (like it does now).
1. Retrieving all preferred reviewers for the file paths included within a PR. Since the pool of dedicated docs reviewers is small, the workflow bot gets all preferred reviewers for a file path, not just one.
1. Merging the two lists of reviewers.

The alternative would be to allow preferred file paths in the docs reviewer pool, but it makes sense to have a small, consistent set of docs reviewers who have experience with docs reviews and a larger pool of subject matter experts for specific file paths from the code reviewer pool.

Implementation notes:

- Adds an `*Assignments.getAllPreferredReviewers` method, which fetches all preferred reviewers for all files in a change set, rather than limiting the number of reviewers per file to 1 as in `getPreferredReviewers`. This is necessary for `CheckInternal` to check that code reviewers with expertise over specific docs paths are counted as valid docs reviewers.
- Changes `getDocsReviewers`, `CheckInternal`, and `checkInternalDocsReviews` to take a slice of `PullRequestFile`s so they can apply the `getPreferredReviewers` logic.

### DIFF
--- a/bot/internal/review/review.go
+++ b/bot/internal/review/review.go
@@ -210,6 +210,7 @@ func (r *Assignments) getDocsReviewers(e *env.Environment, files []github.PullRe
 
 	// If no docs reviewers were assigned, assign admin reviews.
 	if len(reviewers) == 0 {
+		log.Println("No docs reviewers found. Assigning admin reviewers.")
 		return r.getAdminReviewers(e.Author)
 	}
 	return reviewers
@@ -270,14 +271,26 @@ func (r *Assignments) getPreferredReviewers(set []string, files []github.PullReq
 
 // getAllPreferredReviewers returns a list of reviewers that would be
 // preferrable to review the provided changeset. Includes all preferred
-// reviewers for each file path in the chagne set.
+// reviewers for each file path in the changeset.
 func (r *Assignments) getAllPreferredReviewers(set []string, files []github.PullRequestFile) (preferredReviewers []string) {
+	// Check each key in the preferred reviewer map, which is a file path
+	// that reviewers are assigned to. For any file names in the changeset
+	// that begin with that file path, add the reviewers for that pile path
+	// to the set of preferred reviewers. Look up each reviewer in a map to
+	// avoid duplication.
+	assigned := make(map[string]struct{})
 	for path, reviewers := range r.getPreferredReviewersMap(set) {
 		for _, file := range files {
 			if !strings.HasPrefix(file.Name, path) {
 				continue
 			}
-			preferredReviewers = append(preferredReviewers, reviewers...)
+			for _, rev := range reviewers {
+				if _, ok := assigned[rev]; ok {
+					continue
+				}
+				assigned[rev] = struct{}{}
+				preferredReviewers = append(preferredReviewers, rev)
+			}
 		}
 	}
 	return preferredReviewers
--- a/bot/internal/review/review.go
+++ b/bot/internal/review/review.go
@@ -176,14 +176,14 @@ func (r *Assignments) Get(e *env.Environment, changes env.Changes, files []githu
 	switch {
 	case changes.Docs && changes.Code:
 		log.Printf("Assign: Found docs and code changes.")
-		reviewers = append(reviewers, r.getDocsReviewers(e.Author)...)
+		reviewers = append(reviewers, r.getDocsReviewers(e, files)...)
 		reviewers = append(reviewers, r.getCodeReviewers(e, files)...)
 	case !changes.Docs && changes.Code:
 		log.Printf("Assign: Found code changes.")
 		reviewers = append(reviewers, r.getCodeReviewers(e, files)...)
 	case changes.Docs && !changes.Code:
 		log.Printf("Assign: Found docs changes.")
-		reviewers = append(reviewers, r.getDocsReviewers(e.Author)...)
+		reviewers = append(reviewers, r.getDocsReviewers(e, files)...)
 	// Strange state, an empty commit? Return admin reviewers.
 	case !changes.Docs && !changes.Code:
 		log.Printf("Assign: Found no docs or code changes.")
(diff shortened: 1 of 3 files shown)


ptgott:
For context, this `github-terraform` PR adds preferred reviewers for docs, but it looks like we need to edit the workflow bot for this change to take effect: https://github.com/gravitational/github-terraform/pull/694/files

bot/internal/review/review.go
```diff
@@ -260,6 +268,21 @@ func (r *Assignments) getPreferredReviewers(set []string, files []github.PullReq
 	return preferredReviewers
 }
 
+// getAllPreferredReviewers returns a list of reviewers that would be
+// preferrable to review the provided changeset. Includes all preferred
+// reviewers for each file path in the chagne set.
```
zmb3, line 273:
```suggestion
// reviewers for each file path in the change set.
```

bot/internal/review/review.go
```diff
@@ -260,6 +268,21 @@ func (r *Assignments) getPreferredReviewers(set []string, files []github.PullReq
...
+func (r *Assignments) getAllPreferredReviewers(set []string, files []github.PullRequestFile) (preferredReviewers []string) {
+	for path, reviewers := range r.getPreferredReviewersMap(set) {
+		for _, file := range files {
+			if !strings.HasPrefix(file.Name, path) {
+				continue
+			}
+			preferredReviewers = append(preferredReviewers, reviewers...)
```
zmb3, line 280: Feels like this can result in the same reviewer being added multiple times - is this okay?

bot/internal/review/review.go
```diff
@@ -367,7 +390,7 @@ func (r *Assignments) CheckInternal(e *env.Environment, reviews []github.Review,
 		}
 	case changes.Docs && !changes.Code:
 		log.Printf("Check: Found docs changes.")
-		if err := r.checkInternalDocsReviews(e.Author, reviews); err != nil {
+		if err := r.checkInternalDocsReviews(e, reviews, files); err != nil {
```
zmb3, line 393: It's been a while since I've looked at this code and I'm pretty jet-lagged so forgive me if I'm missing something obvious - but why do we need to pass the set of files in for docs reviewers here but not for code reviewers?

bot/internal/review/review.go
```diff
@@ -197,13 +197,20 @@ func (r *Assignments) getReleaseReviewers() []string {
...
-func (r *Assignments) getDocsReviewers(author string) []string {
-	setA, setB := getReviewerSets(author, "Core", r.c.DocsReviewers, r.c.DocsReviewersOmit)
-	reviewers := append(setA, setB...)
+func (r *Assignments) getDocsReviewers(e *env.Environment, files []github.PullRequestFile) []string {
+	// See if any code reviewers are designated preferred reviewers for one of
+	// the changed docs files. If so, add them as docs reviewers.
+	a, b := getReviewerSets(e.Author, "Core", r.c.CodeReviewers, r.c.CodeReviewersOmit)
+	prefCodeReviewers := r.getAllPreferredReviewers(append(a, b...), files)
+
+	// Get the docs reviewer pool, which does not depend on the files
```
zmb3, line 206: Can we skip this step if we found enough preferred reviewers in the previous step?

bot/internal/review/review.go
```diff
@@ -197,13 +197,20 @@ func (r *Assignments) getReleaseReviewers() []string {
...
+	// changed by a pull request.
+	docsA, docsB := getReviewerSets(e.Author, "Core", r.c.DocsReviewers, r.c.DocsReviewersOmit)
+	reviewers := append(prefCodeReviewers, append(docsA, docsB...)...)
 
 	// If no docs reviewers were assigned, assign admin reviews.
 	if len(reviewers) == 0 {
-		return r.getAdminReviewers(author)
+		return r.getAdminReviewers(e.Author)
```
zmb3, line 213: I know this isn't related to your changes, but while we're here I think it would be nice to log cases like this where we fall back to admin reviewers.

bot/internal/review/review.go
ptgott: Oversight on my part! I've changed this to deduplicate `preferredReviewers`. The triple `for` loop isn't the greatest (it assumes the number of file paths and reviewers per path remains small), so I've explained the logic in a comment.

bot/internal/review/review.go
ptgott: That's a good question. For me, it'd be good to keep the reviewer pool wide to increase the likelihood that someone is available to review a docs PR, but that could also mean unnecessary GitHub notification spam. I'm happy to make this change if you think it makes sense. How many reviewers is enough?

bot/internal/review/review.go
ptgott: `checkInternalCodeReviews` gets all code reviewers for the team associated with the PR repo, then checks the reviewers who approved the PR for whether there are multiple code owners or one code owner and one reviewer within that team. It doesn't need the PR files, just the team assignments and reviews.

In `checkInternalDocsReviews, whether a reviewer counts as "internal" or not depends on the file paths within the changeset. Code reviewers are normally excluded from docs reviews unless they have `preferredReviewerFor` configured.

bot/internal/review/review.go
zmb3: I'm fine with leaving it as-is.

zmb3 (approved):

PR #180, opened 2023-11-13
Skip Dependency Review action in merge queue

rosstimothy:
Various attempts to skip the action were tried in `teleport.e`, but they all resulted in the Dependency Review check never reporting a status. This blocked all PRs from landing until the merge queue was disabled. By defining the skip here instead of in an upstream repo it should prevent the aforementioned issue.

### DIFF
--- a/.github/workflows/dependency-review-merge-queue.yaml
+++ b/.github/workflows/dependency-review-merge-queue.yaml
@@ -1,0 +1,25 @@
+# This check runs only on PRs that are in the merge queue.
+#
+# PRs in the merge queue have already been approved, but the dependency check
+# is still required, so this workflow allows the required check to succeed,
+# otherwise PRs in the merge queue would be blocked indefinitely.
+#
+# See "Handling skipped but required checks" for more info:
+#
+# https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+#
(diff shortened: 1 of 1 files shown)


jentfoo:
@reedloden I was curious how this was working for `teleport`, and I see we never re-required `Dependency Review`  after disabling in January due to a couple rust issues: https://github.com/gravitational/github-terraform/commit/fe2b03a99c6b

Unfortunately only one of the linked commits has been merged.  Should we require Dependency Review on `e`?  I feel like whatever strategy we do for `e` should be in sync with `teleport`.  Alternatively we could start to require it again on both and just disable the next time we have Rust work.  Just brainstorming, let me know your thoughts.

reedloden:
Can we just add `oid-registry` to the ignored list for now?

As far as merge queue goes, I suspect we need to do something like this:
https://github.com/actions/dependency-review-action/issues/456#issuecomment-1537840047

jakule:
@reedloden @jentfoo Friendly ping

r0mant (approved):

reedloden:
So, if we need to land this to unbreak something, we can, but this is not the correct fix. We need this action to run within the merge queue.

I'd rather us re-enable dependency review everywhere and just explicitly allowlist those rust packages for now.

reedloden:
I just committed https://github.com/gravitational/shared-workflows/pull/183 and https://github.com/gravitational/shared-workflows/pull/184, which I think will help here. I've re-enabled merge queue in `teleport.e`.

PR #182, opened 2023-11-14
Add `WTFPL` as a permitted license

reedloden:
* http://www.wtfpl.net/
* https://en.wikipedia.org/wiki/WTFPL

Note that Google has `WTFPL` in their 'forbidden' list of licenses for open source projects that people should not contribute to. However, we're not contributing here and only using software written under this license.

https://opensource.google/documentation/reference/patching#forbidden

This is being added to address the addition of `@cspell/dict-django`, which uses it.
https://github.com/gravitational/docs/actions/runs/6661528177/job/18461330208?pr=410

### DIFF
--- a/.github/workflows/dependency-review.yaml
+++ b/.github/workflows/dependency-review.yaml
@@ -41,6 +41,6 @@ jobs:
         uses: actions/dependency-review-action@v3
         with:
           fail-on-severity: moderate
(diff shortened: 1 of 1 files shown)


adaadb6 (approved):

jentfoo (commented):
WTFPL does still have some concerns for us:
* It does not describe patent protections, which means that there may be some risk in including WTFPL software
* Combining with other license (like Apache 2) may not be compatible.  I assume we plan to distribute the WTFPL component compiled under our license?

You're right that this is less concerning bringing in these components, however I am not certain that it is safe considering how minimal the license is.  Can you explain the need to bring this license in?

benarent:
Just checking in on this, it's required for https://github.com/gravitational/docs/pull/410 to pass.

reedloden:
Going to land this now to unblock, but will follow-up if we can better scope this down more.

PR #185, opened 2023-11-23
 Add verification of DB migrations for access-graph 

justinas:
Corresponding CI changes in `access-graph`: https://github.com/gravitational/shared-workflows/pull/185 (WIP)

* Adds ability to check for DB migration errors in the `access-graph` repo
* Makes `reviewers` parameter optional when not running `assign` or `check` workflows
* Improves logging, including a stack trace when the workflow exits with an error

### DIFF
--- a/bot/main.go
+++ b/bot/main.go
@@ -137,7 +137,7 @@ func parseFlags() (flags, error) {
 	if *token == "" {
 		return flags{}, trace.BadParameter("token missing")
 	}
-	if *reviewers == "" && !*local {
+	if *reviewers == "" && !*local && (*workflow == "assign" || *workflow == "check") {
 		return flags{}, trace.BadParameter("reviewers required for assign and check")
 	}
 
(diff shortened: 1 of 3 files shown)


bot/main.go
```diff
@@ -178,9 +178,12 @@ func createBot(ctx context.Context, flags flags) (*bot.Bot, error) {
...
 	}
-	reviewer, err := review.FromString(flags.reviewers)
-	if err != nil {
-		return nil, trace.Wrap(err)
+	var reviewer *review.Assignments
+	if flags.workflow == "assign" || flags.workflow == "check" {
+		reviewer, err = review.FromString(flags.reviewers)
+		if err != nil {
+			return nil, trace.Wrap(err)
+		}
```
fheinecke, line 186: I would be very careful about this. This value is used in several other places that are not immediately clear. I took a look into doing the exact same thing a couple of months ago and decided against it because it would break other workflows without much more refactoring.

jimbishopp (approved):

PR #194, opened 2023-12-05
Fixed `dependency-review` check failing on Teleport API

fheinecke:
This check does not support repositories with multiple licenses due to the way the license query logic is implemented upstream. This PR adds an exception for the Teleport API package, which will affect all repos using this artifact.

This syntax is not very pretty, but this is the most simple solution that I could come up with due to GHA expression weirdness.

Tested via https://github.com/gravitational/cloud/pull/6901


fheinecke:
The linter workflow is failing but I'm not going to fix it in this PR as this PR did not introduce the failing change.

r0mant (approved):

PR #195, opened 2023-12-06
Automatically copy changelog message to the backport or put no-changelog label

AntonAM:
With the new requirement of specifying changelog in the PR body or no-changelog label there's a problem with automatic backports, since they automatically fail those requirements. This PR improves backporting procedure by copying changelog message from the parent PR if it's present, or setting no-changelog label.

### DIFF
--- a/bot/internal/bot/backport.go
+++ b/bot/internal/bot/backport.go
@@ -30,9 +30,10 @@ import (
 	"strings"
 	"text/template"
 
-	"github.com/gravitational/shared-workflows/bot/internal/github"
-
 	"github.com/gravitational/trace"
+	"golang.org/x/exp/slices"
+
+	"github.com/gravitational/shared-workflows/bot/internal/github"
 )
 
 // Backport will create backport Pull Requests (if requested) when a Pull
@@ -79,6 +80,11 @@ func (b *Bot) Backport(ctx context.Context) error {
 
 	var rows []row
 
+	g := git
+	if b.c.Git != nil {
+		g = b.c.Git
+	}
+
 	// Loop over all requested backport branches and create backport branch and
 	// GitHub Pull Request.
 	for _, base := range branches {
@@ -92,7 +98,7 @@ func (b *Bot) Backport(ctx context.Context) error {
 			base,
 			pull,
 			head,
-			git,
+			g,
 		)
 		if err != nil {
 			log.Printf("Failed to create backport branch:\n%v\n", trace.DebugReport(err))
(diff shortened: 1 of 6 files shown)


bot/internal/bot/label.go
```diff
@@ -69,7 +80,14 @@ func (b *Bot) labels(ctx context.Context, files []github.PullRequestFile) ([]str
 	// The branch name is unsafe, but here we are simply adding a label.
 	if b.c.Environment.Repository != env.CloudRepo && isReleaseBranch(b.c.Environment.UnsafeBase) {
 		log.Println("Label: Found backport branch.")
-		labels = append(labels, "backport")
+		labels = append(labels, backportLabel)
+
+		// If automatic backport body doesn't have changelog entry it means there shouldn't be one, otherwise
+		// it would be automatically added. Mark PR with no-changelog label.
+		if len(b.getChangelogEntries(prBody)) == 0 && strings.HasPrefix(b.c.Environment.UnsafeHead, botBackportBranchPrefix) {
```
jentfoo, line 87: As an alternative to this, should we just copy all labels (except the `backport` labels).  That way if `no-changelog` (and other attributes) are set on the original PR, they are defaulted to the backport.

jentfoo (approved):

bot/internal/bot/backport.go
```diff
@@ -117,7 +131,7 @@ func (b *Bot) Backport(ctx context.Context) error {
 				RawQuery: url.Values{
 					"expand": []string{"1"},
 					"title":  []string{fmt.Sprintf("[%v] %v", strings.Trim(base, "branch/"), pull.UnsafeTitle)},
-					"body":   []string{fmt.Sprintf("Backport #%v to %v", b.c.Environment.Number, base)},
+					"body":   []string{bodyText},
```
r0mant, line 134: You can just add "labels" here than modifying labels.go, would be a little bit cleaner. I actually had this logic implemented and sitting in https://github.com/gravitational/shared-workflows/pull/148 for a while but never got a chance to wrap it up.

r0mant: https://github.com/gravitational/shared-workflows/pull/148/files#diff-6a92e12adffd18e18a79d45839beff2d7b47b7d29ae8335552c7faa19b83966bR127

bot/internal/bot/label.go
AntonAM: I've removed any changes to the label workflow in this PR, now setting the label on the backport workflow itself, as @r0mant suggested. I think we can actually do copying of all labels on the backport workflow, but I feel like that should be a separate PR if we decide to do that.

bot/internal/bot/backport.go
AntonAM: Changed. I had to force push to overwrite the commit, because during writing tests for backport flow it ran real git commands on my machine and overwrote my git user (it's safe now, dry run in tests).

r0mant (approved):
Thanks @AntonAM! 🙌

fheinecke:
Thanks @AntonAM!

Issue #205, opened 2023-12-26
Single Approvals and Cross-Repo Approvers


jimbishopp:
This issue describes two problems that require upgrading our internal review bot.

### Single Approvers
As our automation has matured, we have increased the number of pull requests required to upgrade our cloud platform. Many of these new pull requests are simple configuration changes - e.g. [bumping an image version](https://github.com/gravitational/cloud/pull/7106) or [changing a component's resource requirements](https://github.com/gravitational/cloud/pull/7150). **Since these PRs fall into our normal approval flow, they each require two approvals.** This requirement increases the number of reviews our engineers are required to perform and increases the time it takes to push a simple configuration change through to production.

To counter this inefficiency, we will upgrade the review bot to allow single approvers for specific repository paths. 

_While there is no audit requirement for two approvers, we do want to continue requiring two approvals for any code or platform change that isn't considered a simple configuration change._

### Cross-Repo Reviewers
Our cloud platform now hosts components that are primarily operated and maintained by core engineers. These components are configured in the cloud repo and deployed by cloud CD. The problem is that all pull requests submitted to the cloud repository are only reviewable by the cloud team. This is terribly inefficient. To resolve, the review bot should allow a single reviewer to review pull requests in both the cloud and core repositories.

### Examples
* The Core TAG team should be able to approve [bumping the TAG version](https://github.com/gravitational/cloud/pull/7111) without asking for reviews from the cloud team
* The Cloud Operations team should be able to [bump an image version](https://github.com/gravitational/cloud/pull/7106) without requiring two approvals.
* The Core Tooling team, who owns the Teleport release server, should be able to [make changes to their service](https://github.com/gravitational/cloud/pull/6972) without requiring a review from the cloud team.

zmb3:
I support this plan and agree that core and cloud should be able to review each others changes in all repos. This has caused friction in a number of places (EUBP team can't approve cloud PRs, cloud growth team can't approve Teleport UI PRs, etc).

jimbishopp:
https://github.com/gravitational/github-terraform/pull/755

PR #206, opened 2023-12-26
bot: Single Approval Paths

jimbishopp:
This PR addresses the "Single Approver" issue described in #205.

Add the ability for certain PRs to require a single approval. 

A new setting has been added, singleApproverPaths, that defines paths in a repository that only require a single approver. If all the files in a PR only match paths defined in singleApproverPaths, then the PR only requires a single approval. Otherwise, the PR requires the default number of approvals, which is currently defined as 2.

A second setting was added, singleApproverAuthors, that defines PR authors that only require a single approval (dependency bots in cloud for now).

### DIFF
--- a/bot/internal/bot/bot.go
+++ b/bot/internal/bot/bot.go
@@ -180,14 +180,9 @@ func approverCount(authors, paths []string, author string, files []github.PullRe
 		return env.DefaultApproverCount
 	}
 	for _, file := range files {
-		var match bool
-		for _, path := range paths {
-			if strings.HasPrefix(file.Name, path) {
-				match = true
-				break
-			}
-		}
-		if !match {
+		if !slices.ContainsFunc(paths, func(path string) bool {
+			return strings.HasPrefix(file.Name, path)
+		}) {
 			return env.DefaultApproverCount
 		}
 	}
--- a/bot/internal/bot/bot.go
+++ b/bot/internal/bot/bot.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"github.com/gravitational/trace"
+	"golang.org/x/exp/slices"
 
 	"github.com/gravitational/shared-workflows/bot/internal/env"
 	"github.com/gravitational/shared-workflows/bot/internal/github"
@@ -170,10 +171,8 @@ func approverCount(authors, paths []string, author string, files []github.PullRe
 	}
 
 	// check if pr author only requires a single approval
-	for _, a := range authors {
-		if author == a {
-			return 1
-		}
+	if slices.Contains(authors, author) {
+		return 1
 	}
 
 	// check if pr files only require a single approval
--- a/bot/internal/bot/bot.go
+++ b/bot/internal/bot/bot.go
@@ -132,12 +132,13 @@ func New(c *Config) (*Bot, error) {
 
 // classifyChanges determines whether the PR contains code changes
 // and/or docs changes.
-func classifyChanges(e *env.Environment, files []github.PullRequestFile) env.Changes {
+func classifyChanges(c *Config, files []github.PullRequestFile) env.Changes {
 	ch := env.Changes{
-		Large:   !e.IsCloudDeployBranch() && xlargeRequiresAdminApproval(files),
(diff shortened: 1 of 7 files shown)


bot/internal/bot/bot.go
```diff
@@ -153,6 +154,28 @@ func classifyChanges(e *env.Environment, files []github.PullRequestFile) env.Cha
...
+// in the PR against a set of paths that only require a single approver. 1 is returned when all of the
+// files match a single approver path, otherwise env.DefaultApproverCount is returned.
+func approverCount(singleApproverPaths []string, files []github.PullRequestFile) int {
+	if len(singleApproverPaths) == 0 || len(files) == 0 {
+		return env.DefaultApproverCount
+	}
+	for _, file := range files {
+		var match bool
+		for _, path := range singleApproverPaths {
+			if strings.HasPrefix(file.Name, path) {
```
jentfoo, line 167: Currently this will match partial filenames, for example a allowed prefix of `src/dir` would also match to `src/dir2`.  Maybe not concerning, but I wanted to mention it.

jentfoo (approved):

bot/internal/bot/bot.go
jimbishopp: Thanks. This is by design and is the same as what we do for preferred reviewer paths.

jimbishopp:
I'm thinking it may be better to remove the single approver paths from the config file and just define it as a go literal.

jimbishopp:
I added another commit that allows certain authors such as dependabot or renovate to only require a single approver.

bot/internal/bot/bot.go
```diff
@@ -153,6 +159,42 @@ func classifyChanges(e *env.Environment, files []github.PullRequestFile) env.Cha
...
+	if len(files) == 0 {
+		return env.DefaultApproverCount
+	}
+
+	// check if pr author only requires a single approval
+	for _, a := range authors {
+		if author == a {
+			return 1
+		}
+	}
```
zmb3, line 177:
```suggestion
	if slices.Contains(authors, author) {
		return 1
	}
```

bot/internal/bot/bot.go
```diff
@@ -153,6 +159,42 @@ func classifyChanges(e *env.Environment, files []github.PullRequestFile) env.Cha
 	return ch
 }
 
+// approverCount returns the number of required approvers for the PR by comparing the files included
+// in the PR against a set of paths that only require a single approver and the PR author against
+// a set of authors that only require a single approver. 1 is returned when all of the
+// files match a single approver path or the PR author matches one of the authors, otherwise
+// env.DefaultApproverCount is returned.
+func approverCount(authors, paths []string, author string, files []github.PullRequestFile) int {
```
zmb3, line 167: Should we enforce some sort of naming convention for bots that only require a single approval as an extra precaution? I don't think we'd ever want an employee to unconditionally only require a single approval.

Alternatively, we can just require that any author configured as a single approval also passes `isAllowedRobot`.

bot/internal/bot/bot.go
```diff
@@ -153,6 +159,42 @@ func classifyChanges(e *env.Environment, files []github.PullRequestFile) env.Cha
...
+		}
+	}
+
+	// check if pr files only require a single approval
+	if len(paths) == 0 {
+		return env.DefaultApproverCount
+	}
+	for _, file := range files {
+		var match bool
+		for _, path := range paths {
```
zmb3, line 185: nit: `slices.ContainsFunc` might clean this loop up.

zmb3 (approved):

bot/internal/bot/bot.go
jimbishopp: My muscle memory hasn't gotten used to the slices package yet. I need to code more :/

bot/internal/bot/bot.go
jimbishopp: Good idea. I added a test to ensure only bots can be configured as single approver authors. We can also export `isAllowedRobot` from the `review` package and test in `approverCount` if you think we should have a secondary set of checks.

PR #207, opened 2023-12-26
bot: Allow the same reviewer in both core and cloud repos

jimbishopp:
This PR addresses the "Cross-Repo Reviewers" issue described in https://github.com/gravitational/shared-workflows/issues/205.

The current bot configuration data model doesn't allow an engineer to be a reviewer in both the cloud and core repositories. This is an issue because we now require the core team to submit PRs to the cloud repo to manage either tools they own (e.g. release server) or deploy a new version of a component they own (e.g. TAG). 

This PR provides a hacky solution that will allow us to change the bot config data model to include a cloud and core team section, while keeping the original code reviewers map in the bot so we don't have to change a large amount of code.

### DIFF
--- a/bot/internal/bot/bot.go
+++ b/bot/internal/bot/bot.go
@@ -18,10 +18,10 @@ package bot
 
 import (
 	"context"
+	"slices"
 	"strings"
 
 	"github.com/gravitational/trace"
-	"golang.org/x/exp/slices"
 
 	"github.com/gravitational/shared-workflows/bot/internal/env"
 	"github.com/gravitational/shared-workflows/bot/internal/github"
--- a/bot/internal/env/env.go
+++ b/bot/internal/env/env.go
@@ -132,6 +132,14 @@ func (e *Environment) IsCloudDeployBranch() bool {
 		(e.UnsafeBase == CloudProdBranch || e.UnsafeBase == CloudStagingBranch)
 }
 
+// Team returns CloudTeam when the repository is the cloud repo otherwise CoreTeam.
+func (e *Environment) Team() string {
+	if e.Repository == CloudRepo {
+		return CloudTeam
+	}
+	return CoreTeam
+}
+
 func readEvent() (*Event, error) {
 	f, err := os.Open(os.Getenv(githubEventPath))
 	if err != nil {
--- a/bot/internal/env/env.go
+++ b/bot/internal/env/env.go
@@ -132,8 +132,8 @@ func (e *Environment) IsCloudDeployBranch() bool {
 		(e.UnsafeBase == CloudProdBranch || e.UnsafeBase == CloudStagingBranch)
 }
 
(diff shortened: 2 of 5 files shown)


bot/internal/review/review.go
```diff
@@ -109,6 +109,8 @@ type Config struct {
 	// CodeReviewers and CodeReviewersOmit is a map of code reviews and code
 	// reviewers to omit.
 	CodeReviewers     map[string]Reviewer `json:"codeReviewers"`
+	CoreReviewers     map[string]Reviewer `json:"coreReviewers"`
+	CloudReviewers    map[string]Reviewer `json:"cloudReviewers"`
```
zmb3, line 113: nit: keep `CodeReviewers` and `CodeReviewersOmit` together so the comment above still applies to both of them.

(then you can add a separate comment that describes how `CoreReviewers` and `CloudReviewers` work)

bot/internal/env/env.go
```diff
@@ -132,6 +132,14 @@ func (e *Environment) IsCloudDeployBranch() bool {
 		(e.UnsafeBase == CloudProdBranch || e.UnsafeBase == CloudStagingBranch)
 }
 
+// Team returns CloudTeam when the repository is the cloud repo otherwise CoreTeam.
+func (e *Environment) Team() string {
```
zmb3, line 136:
```suggestion
func (e *Environment) RepoOwner() string {
```

Perhaps?

zmb3 (commented):
Can you include an example of what the new reviewers config will look like? I'm having a hard time telling exactly how this changes things.

jimbishopp:
> Can you include an example of what the new reviewers config will look like? I'm having a hard time telling exactly how this changes things.

@zmb3 the test was updated to demonstrate what the new config will look like. Essentially `codeReviewers` will be replaced by `coreReviewers` and `cloudReviewers`. I didn't remove the `Team` attribute because it is baked into so much logic but it could either be removed or set when loading the config file. I think eventually we'll need to have a data model that defined reviewers per repo but I didn't want to take that on right now.

https://github.com/gravitational/shared-workflows/blob/d85387b594e8543336afa9a6f9107a886c492af9/bot/internal/review/review_test.go#L1196-L1217

bot/internal/env/env.go
jimbishopp: Changed to `RepoOwnerTeam` since it returns a `Team` name.

jimbishopp:
I added commit that removes the need to set the `Team` field in the `reviewers.json` configuration file. It is now set automatically when the file is loaded.

bot/internal/bot/bot.go
```diff
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"github.com/gravitational/trace"
+	"golang.org/x/exp/slices"
```
zmb3, line 24:
```suggestion
	"slices"
```

zmb3 (approved):

PR #211, opened 2024-01-08
Add shared migration path to AccessGraphRepo

jakule:
Recently, Access Graph added a `shared` DB schema with their own migrations. This change adds the new path to the DB migrations verification logic.


justinas (approved):

wadells (commented):
Is this still needed?

PR #212, opened 2024-01-08
bot: Fix cloud reviewer assignments

jimbishopp:
Fix an issue with cloud repo assignments, remove the Team field, and add a `preferredOnly` flag.

Context: our reviewer data model was changed last month to allow core team members to review code in the cloud repository. The new team maps (CoreReviewers and CloudReviewers) can include both core and cloud engineers.

1. **Fix Issue:** core team members submitting changes to the cloud repo were treated as external contributors because the internal check when getting reviewers was not checking the new team maps  (CoreReviewers & CloudReviewers).
2. **Remove the Team field:** since we’ve moved team assignments (core vs cloud vs docs) to distinct maps, this team field is not longer required and keeping it created ambiguity. It was previously used to determine which team (core vs cloud vs internal) a reviewer belonged to but team assignments are now determined by which map a reviewer is included in. 
3. **Remove CodeReviewers map:** I kept this old assignment map in the code in an attempt to keep the changes small when making the previous change to allow both core and cloud reviewers to review cloud changes but this turned out to be a bad decision because it introduced a bug where core engineers were considered external users when contributing to the cloud repo.
4. **Add PreferredOnly field to Reviewer struct:** when true, a reviewer is **only** included in preferred reviewers and not the standard reviewer logic. This allows core reviewers in cloud to **only** be added as a reviewer for preferred reviewer paths.

**Background:**
This PR aligns the reviewer code with the `reviewers.json` data model. The data model has a number of related entities: teams, repositories, reviewers, and paths. Ideally, the top-level reviewer groups would be the repository since a preferredReviewerPath's context is a specific repository. Our data model adds a bit of ambiguity by defining teams (core or cloud), where the cloud team manages the cloud repository and the core team manages the teleport and teleport.e repositories. This change does not address the repository ambiguity but does allow for a single reviewer to be defined in both cloud and core teams.

### DIFF
--- a/bot/internal/review/review.go
+++ b/bot/internal/review/review.go
@@ -94,6 +94,8 @@ func isAllowedRobot(author string) bool {
 type Reviewer struct {
 	// Owner is true if the reviewer is a code or docs owner (required for all reviews).
 	Owner bool `json:"owner"`
+	// PreferredOnly is true if the reviewer should only be included in preferred reviewer paths.
+	PreferredOnly bool `json:"preferredOnly",omitempty`
 	// PreferredReviewerFor contains a list of file paths that this reviewer
 	// should be selected to review.
 	PreferredReviewerFor []string `json:"preferredReviewerFor,omitempty"`
@@ -286,10 +288,14 @@ func (r *Assignments) getCodeReviewers(e *env.Environment, files []github.PullRe
 	// pick from the overall set at random.
 	resultingSetA := preferredSetA
 	if len(resultingSetA) == 0 {
+		// Remove reviewers from setA whose preferredOnly field is false.
+		setA = filterPreferredOnly(reviewers, setA, false)
 		resultingSetA = append(resultingSetA, setA[r.c.Rand.Intn(len(setA))])
 	}
 	resultingSetB := preferredSetB
 	if len(resultingSetB) == 0 {
+		// Remove reviewers from setB whose preferredOnly field is false.
+		setB = filterPreferredOnly(reviewers, setB, false)
 		resultingSetB = append(resultingSetB, setB[r.c.Rand.Intn(len(setB))])
 	}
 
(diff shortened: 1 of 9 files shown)


bot/internal/review/review.go
```diff
@@ -218,9 +207,10 @@ func (r *Assignments) IsInternal(author string) bool {
 		return true
 	}
 
-	_, code := r.c.CodeReviewers[author]
+	_, core := r.c.CoreReviewers[author]
+	_, cloud := r.c.CloudReviewers[author]
 	_, docs := r.c.DocsReviewers[author]
-	return code || docs
+	return core || cloud || docs
```
jimbishopp, line 213: This is the main change. The rest is removing the team field and fixing tests.

bot/internal/review/review.go
```diff
@@ -385,24 +375,12 @@ func (r *Assignments) getAdminReviewers(author string) []string {
 func (r *Assignments) getCodeReviewerSets(e *env.Environment) ([]string, []string) {
 	// Internal non-Core contributors get assigned from the admin reviewer set.
 	// Admins will review, triage, and re-assign.
-	v, ok := r.c.CodeReviewers[e.Author]
-	if !ok || v.Team == env.InternalTeam {
+	if !r.IsInternal(e.Author) {
```
jimbishopp, line 378: Main change part 2: calling IsInternal rather than checking only CodeReviewers and the Team field

bot/internal/review/review.go
```diff
@@ -385,24 +374,12 @@ func (r *Assignments) getAdminReviewers(author string) []string {
...
```
jimbishopp, line 377: This change resolves the issue where a core engineer not defined in the cloud reviewer map submits a PR to the cloud repo but requires admin approval. Calling IsInternal now checks core, docs, **and cloud** maps to determine if the author is internal.

zmb3 (approved):

PR #214, opened 2024-01-11
Added AMI cleanup tool

fheinecke:
This PR adds a tool to cleanup old dev tag AMIs. Last time this was ran manually it reduced our annual AWS spend by about $400k.

This PR is the first one for this repo that conforms to [RFD 1](https://github.com/gravitational/shared-workflows/blob/main/rfd/0001-tooling-requirements.md). As such, a lot of the changes are repo-wide tooling needed to meet these requirements. These features will make it much easier to add future projects to this repo.

The following features have been added:
* The original [Python script](https://gravitational.slack.com/archives/C0351AUPJ8K/p1695715251834689?thread_ts=1695708587.693479&cid=C0351AUPJ8K) that @reedloden has been rewritten in Go.
* Wrote unit tests covering most of the program's logic.
* Wrote tooling to build multi-arch binaries, tarballs, and container images. See buddy PR https://github.com/gravitational/shared-workflows/pull/239.
* Wrote PR and publishing workflows. See buddy PR https://github.com/gravitational/shared-workflows/pull/239.
* Added a GitHub Action (not workflow) to use this tool via GHA.
* Setup and configured Renovate to manage this project only. See buddy PR https://github.com/gravitational/shared-workflows/pull/237.

Post merge, the following work needs to be done:
* The GHA environment for this repo needs to be configured for Renovate.
* The action needs to be used somewhere (probably teleport.e) to actually prune images.

The tool has been tested locally on one of the dev accounts, but the rest of this (GHA workflows, project-specific Renovate configuration) have not been. I expect that there are some problems here that I won't be able to fix until the PR lands and they actually run.

Updates gravitational/teleport.e#3187


tools/ami-cleanup/workflows/cd.yaml
```diff
@@ -0,0 +1 @@
+../../../.github/workflows/ami-cleanup-cd.yaml
```
wadells, line 1: Should these symlinks go the other way?  The originals in `tools/` and the symlink in `.github/`?

https://github.com/gravitational/shared-workflows/blob/d8562e07c65a15c5da4472f6d1601f5ea6e9bedd/rfd/0001-tooling-requirements.md#L61-L62

wadells (commented):
My big question:

Did you run earthly by @camscale, @tcsc and @r0mant?

From what I've read of it, I'm partial to the tool for isolation and reproducibility -- not to mention "better than the 1980s Makefile syntax".  However it is a notable new tool that isn't in our toolchain yet.

Other feedback:
* I might avoid adding "ami" to the name of this tool.  I think it is likely we'll see cleanup workflows expand to other artifact types.  Knowing how things go, there is a good chance that e.g. S3 or ECR cleanup could get added to this tool.
* Including renovate in the same PR is a lot.  I might recommend we include it in a separate PR.  It looks like it is mostly for Earthfiles currently, but it'd be good to discuss the long term goal for renovate vs dependabot in this repo, and I'm not sure we want to sidtrack the cleanup script discussion with that.

.github/renovate.json5
```diff
@@ -0,0 +1,63 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:recommended",
+    ":enableRenovate",
+    ":semanticCommits",
+    // TODO enable these after extensive testing
```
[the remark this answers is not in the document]
fheinecke: Removed that and another related section. This is leftover from when this discussion came up in cloud-terraform. I copy/pasted the file here, and just looked over that comment.

.github/renovate.json5
```diff
@@ -0,0 +1,63 @@
...
+    // These predate RFD 0001 and are managed by dependabot
+    "bot/**", 
+    ".github/workflows/codeql.yml",
+    ".github/workflows/csv-lint.yaml",
+    ".github/workflows/dependency-review.yaml",
+    ".github/workflows/github-action-lint.yaml",
+    ".github/workflows/govulncheck.yaml",
+    ".github/workflows/json-lint.yaml",
+    ".github/workflows/terraform-lint.yaml",
+    ".github/workflows/trivy.yaml"
```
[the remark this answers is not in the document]
fheinecke: > I'm fine swapping all these to renovate

I'll put this on my todo list. I don't really want to include it in this PR to avoid scope creep. Unfortunately Dependabot is a non-option for a lot of things due to how rigid it is (compared to renovate).

.github/workflows/ami-cleanup-ci.yaml
```diff
@@ -0,0 +1,43 @@
...
+      pull-requests: read
+    outputs:
+      should-verify-pr: ${{ steps.filter.outputs.changed }} # True if the AMI cleanup tool changed, false otherwise
+    steps:
+      - name: Filter out unrelated changes
+        uses: dorny/paths-filter@4512585405083f25c027a35db413c2b3b9006d50 # v2.11.1
+        id: filter
+        with:
+          filters: |
+            changed: "tools/ami-cleanup/**"
```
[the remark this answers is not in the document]
fheinecke: I _think_ this will be caught, but I don't have an easy way to test in this PR. There is a symlink in the workflows directory which (hopefully) renovate will follow.

tools/ami-cleanup/LICENSE
```diff
@@ -0,0 +1,662 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
```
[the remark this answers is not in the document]
fheinecke: That works for me. I emailed the open source group (per our company policy) several weeks ago about this and never got a response back, so I just went off of what we're doing for Teleport.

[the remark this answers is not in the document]
fheinecke: Do I need to do anything about setting up a CLA/checks for it?

tools/ami-cleanup/workflows/cd.yaml
fheinecke: Unfortunately no. I just had a discussion with Github support about this (after the workflows initially failed to run) and GHA doesn't support symlinks in the workflows directory. I got a very noncommittal "we'll put it in a file somewhere and think about it" type response when I asked about adding this support.

I realize that this is in contradiction to the RFD, but unfortunately it has to be this way if we want any sort of direct workflow/project linkage.

wadells: Thats fine.  Maybe we should update the RFD then?

fheinecke:
> Did you run earthly by @camscale, @tcsc and @r0mant?

I talked with @r0mant about it briefly during our 1x1 last week. While using it here specifically was not discussed, I figured it would be a good starting place to evaluate the tool. If we don't like it, then only ~150 lines of code are lost (not a huge deal IMO).

I personally think there are some huge benefits to using it over our more traditional makefile/dockerfile/shell script/etc. spagetti. I can go into more details here if desired, but I put [this item](https://gravitational.slab.com/posts/proposed-teleport-core-goals-fy-25-q-1-vt3xq0ky#hhj54-internal-tooling:~:text=Evaluate%20overall%20build%20process%20complexity%20to%20discover%20specific%20work%20for%20future%20quarters%20(cost%2C%20toil%2C%20maintenance%2C%20risk%20reductions)) as a proposed quarterly goal for our team which should cover discussing the merits and flaws of the solution.

> I might avoid adding "ami" to the name of this tool. I think it is likely we'll see cleanup workflows expand to other artifact types. Knowing how things go, there is a good chance that e.g. S3 or ECR cleanup could get added to this tool.

I thought about this, and I was even thinking about adding public ECR cleanup (ECR lifecycles not supported here) next. However I thought that it might be better to pull common code out into a lib and build a different tool, instead of adding ECR support to this one. Personally I am a fan of the "every tool does one specific job" Unix philosophy (paraphrasing). 

What do you think?

> Including renovate in the same PR is a lot. I might recommend we include it in a separate PR. It looks like it is mostly for Earthfiles currently, but it'd be good to discuss the long term goal for renovate vs dependabot in this repo, and I'm not sure we want to sidtrack the cleanup script discussion with that.

I thought about this pretty heavily, and I'm still not sure which way is better. On one hand:
* All the code (workflow) and 90% of the config is already written/tested/in use, as it was copy/pasted from cloud-terraform
* Renovate is basically useless if there is no tool/dependencies to run it against
* It's a little easier for me to manage/merge one PR in this repo rather than several (PRs here usually sit open for quite a while)
* It's needed for this tool per the RFD

On the other hand, separating it out in another PR would:
* Allow for a more dependency management specific discussion about these changes
* Make this PR a little easier to review

Do you (or other reviewers) have a strong preference one way of another?

tools/ami-cleanup/workflows/cd.yaml
fheinecke: Yea I was thinking about that. I don't think we often update RFDs but I think it is reasonable to do so in this case.

camscale:
> Do you (or other reviewers) have a strong preference one way of another?

Anything you can do to break up a +2,500 line PR is appreciated. If the renovate stuff is essentially independent of the functionality of the tool to clean up the AMIs, then I suggest splitting it out. Layering PRs to build up from a small essential kernel and adding new orthogonal concerns is much easier to review than all in one big bang. At first blush, the bullet point list at the top of the PR description listing the features looks like they could almost be separate PRs too. It may mean adding stuff that does not initially conform to the RFD, but as long as you are heading there I think that's fine.

go.work
zmb3, line 1: go.work is not typically committed to source control.

tools/ami-cleanup/Earthfile
zmb3, line 1: Not sure how I feel about requiring everyone to learn yet another DSL..

Do we need a complicated 300-line build process for this? If it's pure Go, can't we just periodically `go run` it in the `golang:latest` container image and call it a day?

tools/ami-cleanup/src/cmd/main.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/gravitational/shared-workflows/tools/ami-cleanup/internal"
+)
+
+var (
+	Version = "0.0.0-dev"
+)
+
+func main() {
+	kingpin.Version(Version)
+	dryRun := kingpin.Flag("dry-run", "set to only report the expected changes without actually performing them").Bool()
```
zmb3, line 34: Do we really need kingpin for a program that takes a single flag? Less dependencies means less things to patch and keep up to date. Would the `flag` package suffice here?

tools/ami-cleanup/src/cmd/main_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"golang.org/x/mod/semver"
+)
+
+func TestVersion(t *testing.T) {
+	require.True(t, semver.IsValid("v"+Version), "the version string is not a valid semver")
```
zmb3, line 27: I'm not sure we need to write tests for other people's code. I assume being under the golang.org umbrella that this project has enough test coverage as-is.

tools/ami-cleanup/src/internal/aws.go
```diff
@@ -0,0 +1,131 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// TODO move this under /libs at some point, however it is nowhere near
+// large enough to justify separate module today
+package internal
```
zmb3, line 19: `internal` is not a very good package name - it doesn't say anything about what the code in the package does.

https://go.dev/blog/package-names

> Good package names make code better. A package’s name provides context for its contents, making it easier for clients to understand what the package is for and how to use it. The name also helps package maintainers determine what does and does not belong in the package as it evolves. Well-named packages make it easier to find the code you need.

tools/ami-cleanup/src/internal/aws.go
```diff
@@ -0,0 +1,131 @@
...
+// github.com/gravitational/cloud/pkg/aws
+// TODO consider writing a `go generate` program for this. This will quickly get out
+// of hand if used frequently.
+
+// region Account API
+type IAccountApi interface {
+	ListRegions(ctx context.Context, params *account.ListRegionsInput, optFns ...func(*account.Options)) (*account.ListRegionsOutput, error)
+}
+
+type AccountApi struct {
```
zmb3, line 63:
```suggestion
type AccountAPI struct {
```

https://go.dev/wiki/CodeReviewComments#initialisms

tools/ami-cleanup/src/internal/aws.go
```diff
@@ -0,0 +1,131 @@
...
+	return false
+}
+
+// These are primarily used for mocks while testing, and follows a similar pattern to
+// github.com/gravitational/cloud/pkg/aws
+// TODO consider writing a `go generate` program for this. This will quickly get out
+// of hand if used frequently.
+
+// region Account API
+type IAccountApi interface {
```
zmb3, line 59: The `I` prefix is not very idiomatic in Go. I'd remove it and come up with a better name.

That said, I'm not sure you need an interface at all - doesn't aws-sdk-go already include interfaces and mocks?

tools/ami-cleanup/src/internal/aws_test.go
```diff
@@ -0,0 +1,68 @@
...
+						},
+					},
+				},
+			},
+			expectedResult: true,
+		},
+	}
+
+	for _, test := range tests {
+		testCap := test // Capture the loop var, unnecessary in upcoming go 1.22
```
zmb3, line 61: I would remove this and remove the `t.Parallel()`. These tests are pure computation and don't make any syscalls, network calls, etc. Parallelizing it probably doesn't result in any meaningful improvement, and may even slow things down.

tools/ami-cleanup/src/internal/cleanup.go
```diff
@@ -0,0 +1,239 @@
...
+
+import (
+	"context"
+	"log"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/account"
+	accountTypes "github.com/aws/aws-sdk-go-v2/service/account/types"
```
zmb3, line 27:
```suggestion
	accounttypes "github.com/aws/aws-sdk-go-v2/service/account/types"
```

All lower case is the convention for package names (and aliases)

tools/ami-cleanup/src/internal/cleanup.go
```diff
@@ -0,0 +1,239 @@
...
+		totalImagesDeleted += imagesDeleted
+	}
+
+	log.Printf("Deleted %d GiB of %d images across %d regions", totalSpaceRecovered, totalImagesDeleted, len(enabledRegions))
+	return nil
+}
+
+// Cleans up the images in the given region. Returns the total amount of space recovered, number of images deleted,
+// and any error.
+func (ai ApplicationInstance) cleanupRegion(ctx context.Context, cfg aws.Config, regionName string) (int32, int, error) {
```
zmb3, line 87: Why mix int32 and int here? Do you actually care about the width of an integer here?

tools/ami-cleanup/src/internal/cleanup.go
```diff
@@ -0,0 +1,239 @@
...
+				return nil, nil, trace.Wrap(err, "failed to request enabled regions")
+			}
+
+			return results.NextToken, results.Regions, nil
+		},
+	)
+}
+
+func (ai *ApplicationInstance) getDevImagesInRegion(ctx context.Context, client IEc2Api) ([]ec2Types.Image, error) {
+	// These are a weird language workaround for getting a pointer to a boolean literal
```
zmb3, line 139: I think this is what `aws.BoolValue(true)` is for. (And `aws.StringValue`)

tools/ami-cleanup/src/internal/cleanup.go
```diff
@@ -0,0 +1,239 @@
...
+		results = append(results, newResults...)
+		if nextToken == nil {
+			return results, nil
+		}
+		previousToken = nextToken
+	}
+}
+
+func (ai *ApplicationInstance) cleanupImageIfOld(ctx context.Context, client IEc2Api, image ec2Types.Image) (int32, error) {
+	creationDate, err := iso8601.ParseString(*image.CreationDate)
```
zmb3, line 193: Do we need a 3rd party dependency for this? Can `time.Parse` do this for you?

tools/ami-cleanup/Earthfile
fheinecke: We need at minimum the following across PR workflows, release workflows, and local development:
* Binary building (local, PR, release)
* Container image building (local, PR, release)
* Linting (local, PR)
* Test running (local, PR)
* Artifact cleanup (local)
* Publishing the container image, changelog (local, release)

Ideally we'd also publish tarballs so that this can be ran locally (I'm guessing we'll run this locally until it's actually deployed somewhere).

This build tooling allows us to do all of the above with less than 300 lines including comments. On top of that, it runs the exact same thing locally as we do in our CI/CD pipelines - something we've struggled with for Teleport builds. Builds are also completely encapsulated and should be reproducible on any machine that can run Earthly.

I can implement this with our usual makefiles/dockerfiles/shell script approach, but I can promise that it will be significantly more code than this approach.

zmb3: > I can implement this with our usual makefiles/dockerfiles/shell script approach, but I can promise that it will be significantly more code than this approach.

I don't doubt that. I'm just not convinced that we need to build containers, publish artifacts and clean up artifacts. We typically thing of our workflows as scripts that happen to be written in Go instead of bash, and we just compile and run the tip of master.

tools/ami-cleanup/src/internal/cleanup_test.go
```diff
@@ -0,0 +1,667 @@
...
+						MockListRegions: test.mockListRegions,
+					}
+				},
+			}
+
+			// Run the function under test
+			regions, err := application.getEnabledRegions(context.Background(), application.accountClientGenerator(nil))
+
+			// Verify the results
+			checkError(t, test.shouldError, err)
```
zmb3, line 123: Change `shouldError` from a bool to a `require.ErrorAssertionFunc`, and for each test case instead of passing in true/false, pass in `require.Error` or `require.NoError`. I'd rename it from `shouldError` to `checkError` too.

When that's done, you can remove your `checkError` function completely, and just run `test.checkError(t, err)`.

tools/ami-cleanup/Earthfile
r0mant: Yeah, I don't even think it will be more code for just a handful of Makefile targets, it's just a few of go and docker commands, but even if it will, it will be more code that the rest of the team can maintain vs less code that the rest of the team can't understand and has to learn.

tools/ami-cleanup/src/internal/aws.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func NewAccountApi(cfg *aws.Config) IAccountApi {
+	return &AccountApi{
+		cfg:    cfg,
+		Client: account.NewFromConfig(*cfg),
+	}
+}
+
+type MockAccountAPI struct {
```
r0mant, line 75: I would move all objects related to testing into a separate file e.g. `mocks.go` or something similar. Let's not mix "executable" code with purely testing code.

tools/ami-cleanup/src/internal/cleanup.go
```diff
@@ -0,0 +1,239 @@
...
+
+type ApplicationInstance struct {
+	shouldDoDryRun bool
+
+	// Dependency injection for testing
+	accountClientGenerator func(cfg *aws.Config) IAccountApi
+	ec2ClientGenerator     func(cfg *aws.Config) IEc2Api
+}
+
+// Creates a new instance of the tool
```
r0mant, line 42: Let's follow the style guide for godocs: https://go.dev/wiki/CodeReviewComments#comment-sentences.
```suggestion
// NewApplicationInstance creates a new instance of the tool.
```

tools/ami-cleanup/src/internal/cleanup.go
```diff
@@ -0,0 +1,239 @@
...
+func NewApplicationInstance(doDryRun bool) *ApplicationInstance {
+	return &ApplicationInstance{
+		shouldDoDryRun:         doDryRun,
+		accountClientGenerator: NewAccountApi,
+		ec2ClientGenerator:     NewEc2Api,
+	}
+}
+
+// Performs cleanup
```
r0mant, line 51:
```suggestion
// Run performs cleanup ...
```

tools/ami-cleanup/src/internal/cleanup.go
```diff
@@ -0,0 +1,239 @@
...
+
+func (ai *ApplicationInstance) cleanupImageIfOld(ctx context.Context, client IEc2Api, image ec2Types.Image) (int32, error) {
+	creationDate, err := iso8601.ParseString(*image.CreationDate)
+	if err != nil {
+		return 0, trace.Wrap(err, "failed to parse image %q creation timestamp %q as an ISO 8601 value", *image.Name, *image.CreationDate)
+	}
+
+	// If the image is less than a month old, don't do anything
+	imageAge := time.Since(creationDate)
+	if imageAge <= 30*24*time.Hour {
```
r0mant, line 200: Nit: Move the cutoff time into a file-level constant.

tools/ami-cleanup/src/internal/cleanup.go
```diff
@@ -0,0 +1,239 @@
...
+		return 0, 0, trace.Wrap(err, "failed to get a list of dev images for %q", regionName)
+	}
+
+	var totalSpaceRecovered int32
+	totalImagesDeleted := 0
+	totalSnapshotCount := 0
+	for _, devImage := range devImages {
+		imageSpace, err := ai.cleanupImageIfOld(ctx, ec2Client, devImage)
+		if err != nil {
+			return 0, 0, trace.Wrap(err, "failed to cleanup dev image %q", devImage.Name)
```
r0mant, line 104: Should we continue if we failed to cleanup one image instead of returning on first failure?

tools/ami-cleanup/src/internal/cleanup.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+// Performs cleanup
+func (ai *ApplicationInstance) Run(ctx context.Context) error {
+	cfg, err := config.LoadDefaultConfig(ctx)
+	if err != nil {
+		return trace.Wrap(err, "failed to load AWS credentials")
+	}
+
+	enabledRegions, err := ai.getEnabledRegions(ctx, ai.accountClientGenerator(&cfg))
```
r0mant, line 58: Personally I would just allow the tool to accept regions as a CLI argument for simplicity and avoided an extra API call but no strong opinion.

tools/ami-cleanup/src/internal/cleanup.go
```diff
@@ -0,0 +1,239 @@
...
+
+	log.Printf("Deleted %d GiB of %d images across %d regions", totalSpaceRecovered, totalImagesDeleted, len(enabledRegions))
+	return nil
+}
+
+// Cleans up the images in the given region. Returns the total amount of space recovered, number of images deleted,
+// and any error.
+func (ai ApplicationInstance) cleanupRegion(ctx context.Context, cfg aws.Config, regionName string) (int32, int, error) {
+	// A new client must be created for each region
+	cfg.Region = regionName
```
r0mant, line 89: Functions with side effects like this that modify the passed in argument are usually an anti-pattern, can we avoid doing that? E.g. just have a helper method that returns a client for a specified region.

tools/ami-cleanup/src/internal/cleanup.go
```diff
@@ -0,0 +1,239 @@
...
+	ec2Client := ai.ec2ClientGenerator(&cfg)
+
+	log.Printf("Deleting images in %q", regionName)
+	devImages, err := ai.getDevImagesInRegion(ctx, ec2Client)
+	if err != nil {
+		return 0, 0, trace.Wrap(err, "failed to get a list of dev images for %q", regionName)
+	}
+
+	var totalSpaceRecovered int32
```
r0mant, line 98: Nit: Might be helpful to the reader to indicate what is the measure here e.g. "totalBytesRecovered"?

tools/ami-cleanup/src/internal/cleanup.go
```diff
@@ -0,0 +1,239 @@
...
+	if imageAge <= 30*24*time.Hour {
+		return 0, nil
+	}
+
+	log.Printf("\tDeleting %s old AMI %q", imageAge.String(), *image.Name)
+	_, err = client.DeregisterImage(ctx, &ec2.DeregisterImageInput{
+		ImageId: image.ImageId,
+		DryRun:  &ai.shouldDoDryRun,
+	})
+	if err != nil && (!IsDryRunError(err) || !ai.shouldDoDryRun) {
```
r0mant, line 209: Why are we checking for `!ai.shouldDoDryRun` here?

tools/ami-cleanup/src/cmd/main.go
fheinecke: > Do we really need kingpin for a program that takes a single flag?

No, but the last couple of times I brought kingpin vs flag up in slack [the opposite argument was made](https://gravitational.slack.com/archives/C0351AUPJ8K/p1673392009796309?thread_ts=1673053788.975829&cid=C0351AUPJ8K), [or wasn't addressed](https://gravitational.slack.com/archives/C0351AUPJ8K/p1697840072351889).

Do you want me to change this to flag?

tools/ami-cleanup/src/cmd/main_test.go
fheinecke: > I'm not sure we need to write tests for other people's code

That's not what this is testing. This is testing to ensure that `main.Version` is a valid semver. Given that this is overwritten when building via linker flags I thought it would be a good idea to validate the provided value.

tools/ami-cleanup/src/internal/aws.go
fheinecke: This decision was based on the Go standard project layout [here](https://github.com/golang-standards/project-layout/tree/master/internal), which states that for smaller projects extra structure is not required, and that this code can live under a package named `internal`. [One of the examples](https://github.com/influxdata/influxdb/tree/master/internal) provided in the standard uses this naming scheme.

Do you want me to change this? If so, how do you want me to structure the project?

tools/ami-cleanup/src/internal/aws.go
fheinecke: > doesn't aws-sdk-go already include interfaces and mocks?

Unfortunately not in v2. [Cloud uses this same pattern](https://github.com/gravitational/cloud/blob/master/pkg/aws/mocks/mocks.go) (that's where I pulled this from). This is also the approach that [the AWS docs suggest](https://aws.github.io/aws-sdk-go-v2/docs/unit-testing/#mocking-client-operations).

tools/ami-cleanup/src/internal/cleanup.go
fheinecke: _I_ don't, but the AWS API uses int32 for this value. Using int32 keeps the type consistent. Additionally, if I were to switch to `int` via something like `int(<aws API call int32 value>)`, then the value could overflow on 32 bit systems if AWS ever changes the type to `int64` (which would not cause a compilation error and would be unlikely to be manually caught).

tools/ami-cleanup/src/internal/cleanup.go
fheinecke: Thanks - I was not aware of these functions.

tools/ami-cleanup/src/internal/cleanup.go
fheinecke: Strictly speaking, yes. `time` does not fully support iso8601, just a subset. According to the docs the returned value is an iso8601 timestamp string.

tools/ami-cleanup/src/internal/cleanup.go
fheinecke: I think so - if one fails for whatever reason, then it is highly likely that the remaining will fail. The most common failures include invalid AWS credentials, API rate limits, bugs resulting in invalid queries.

tools/ami-cleanup/src/internal/cleanup.go
fheinecke: IMO this is a better approach because:
* If we start publishing images to additional regions (such as when AWS adds additional regions) then we don't have to remember to update some cleanup list somewhere.
* Moving this upwards to a CLI parameter means more shell code (granted this is very little). I'm trying to move as much as reasonably possible via a testable, rather than add additional shell code.

tools/ami-cleanup/src/internal/cleanup.go
fheinecke: Passing the struct by value does create a new copy of the config. See https://go.dev/play/p/yS0-otWZ4KA as an example. The original `cfg` value that is passed upon call to `cleanupRegion` is not modified.

tools/ami-cleanup/src/internal/cleanup.go
fheinecke: The AWS API will return a "dry run error" upon success, which needs special handling to ignore.

fheinecke:
@r0mant and @zmb3 I addressed your review comments. I plan on changing everything I did not comment on, but how I change them may depend on the outcome of your responses to my comments. Please take another look when you get a chance.

tools/ami-cleanup/LICENSE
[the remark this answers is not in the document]
wadells: If we use AGPL, we should set up the CLA using https://cla-assistant.io/.  If we use Apache 2, I don't think we need to do anything.

tools/ami-cleanup/src/cmd/main_test.go
zmb3: Fair point, I'm fine with keeping this here.

We could even rename `TestVersion` to something like `TestTeleportVersionIsSemver`

tools/ami-cleanup/src/cmd/main.go
zmb3: No strong preference, totally up to you.

I now realize I am the one who made the opposite argument last time. Oops!

I suppose that reflects my lack of a strong opinion. I'm okay with whatever you decide.

In my defense, the last PR was in gravitational/teleport where we already have a kingpin dependency. This is a separate repo with a much smaller set of dependencies and also happens to run some of our sensitive workflows, so it seems reasonable to keep dependencies to a minimum.

tools/ami-cleanup/src/internal/aws.go
zmb3: Yeah, I remember when that repo was published there was a whole buzz about it because it is not in any way a "Go standard project layout." This is random internet people who slapped the term "standard" on their project.

I'd be much more inclined to adopt the device in the official Go blog (linked above).

In any case, the point of internal packages is that they reside in the subtree of a directory named internal, not that the package itself is necessarily named internal. This prevents them from being imported outside of the module but still gives you the opportunity to select a package name that better describes its contents.

.vscode/extensions.json
```diff
@@ -0,0 +1,7 @@
+{
+    "recommendations": [
+        "earthly.earthfile-syntax-highlighting",
```
zmb3, line 3: This doesn't seem necessary anymore.

tools/ami-cleanup/README.md
```diff
@@ -0,0 +1,33 @@
...
+Docker: `docker run --rm --env <AWS_CREDENTIAL_VARS> ghcr.io/gravitational/shared-workflows/ami-cleanup:<version> [--dry-run]`
+GHA:
+```yaml
+- uses: gravitational/shared-workflows/tools/ami-cleanup@<version>
+  with:
+    dry-run: <true | false>
+```
+
+## Building:
+This tool requires [Earthly](https://earthly.dev/) to build, used as an alternative to Makefiles. Installation instructions for Earthly are available [here](https://earthly.dev/get-earthly).
```
zmb3, line 18: Remove this?

tools/ami-cleanup/action.yaml
```diff
@@ -0,0 +1,19 @@
+---
+name: Cleanup old AMI images
```
zmb3, line 2:
```suggestion
name: Clean up old AMI images
```

tools/ami-cleanup/src/cleanup/aws.go
```diff
@@ -0,0 +1,96 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// TODO move this under /libs at some point, however it is nowhere near
+// large enough to justify separate module today
+package cleanup
```
zmb3, line 19:
```suggestion
package cleanup
```

In addition to not being necessary, this will also make sure godoc doesn't interpret your TODO as package documentation.

tools/ami-cleanup/src/cleanup/aws.go
```diff
@@ -0,0 +1,96 @@
...
+
+	return false
+}
+
+// These are primarily used for mocks while testing, and follows a similar pattern to
+// github.com/gravitational/cloud/pkg/aws
+// TODO consider writing a `go generate` program for this. This will quickly get out
+// of hand if used frequently.
+
+// region Account API
```
zmb3, line 58:
```suggestion
```

In addition to not being common in Go code, this comment will be interpreted as godoc.

tools/ami-cleanup/src/cleanup/aws.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func NewEC2API(cfg *aws.Config) EC2API {
+	return &AWSEC2API{
+		cfg:    cfg,
+		Client: ec2.NewFromConfig(*cfg),
+	}
+}
+
+// endregion
```
zmb3, line 96: This file is 96 lines and 20% of that is a license. I don't think we need these `region` markers.

tools/ami-cleanup/src/cleanup/aws_test.go
```diff
@@ -0,0 +1,67 @@
...
+						},
+					},
+				},
+			},
+			expectedResult: true,
+		},
+	}
+
+	for _, test := range tests {
+		testCap := test // Capture the loop var, unnecessary in upcoming go 1.22
```
zmb3, line 61: I'd remove this.
1. Your test is not parallel, so it doesn't make a difference.
2. If you must keep it, I'd shadow the test variable `test := test` which would make it impossible to refer to the wrong one.

tools/ami-cleanup/src/cleanup/cleanup.go
```diff
@@ -0,0 +1,237 @@
...
+		totalImagesDeleted += imagesDeleted
+	}
+
+	log.Printf("Deleted %d GiB of %d images across %d regions", totalBytesRecovered, totalImagesDeleted, len(enabledRegions))
+	return nil
+}
+
+// Cleans up the images in the given region. Returns the total amount of space recovered, number of images deleted,
+// and any error.
+func (ai ApplicationInstance) cleanupRegion(ctx context.Context, cfg aws.Config, regionName string) (int32, int, error) {
```
zmb3, line 89:
```suggestion
func (ai *ApplicationInstance) cleanupRegion(ctx context.Context, cfg aws.Config, regionName string) (int32, int, error) {
```

It's a bit odd to have a mix of pointer and non-pointer receivers on the same type.

tools/ami-cleanup/src/cleanup/cleanup.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+}
+
+func (ai *ApplicationInstance) cleanupImageIfOld(ctx context.Context, client EC2API, image ec2types.Image) (int32, error) {
+	creationDate, err := iso8601.ParseString(*image.CreationDate)
+	if err != nil {
+		return 0, trace.Wrap(err, "failed to parse image %q creation timestamp %q as an ISO 8601 value", *image.Name, *image.CreationDate)
+	}
+
+	// If the image is less than a month old, don't do anything
```
zmb3, line 196:
```suggestion
	// don't delete newer images
```

This comment won't go out of date if we change `maxImageKeepAge`

tools/ami-cleanup/src/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,677 @@
...
+				}, nil
+			},
+			expectedRegions: expectedRegions,
+		},
+	}
+
+	for _, test := range tests {
+		if test.checkError == nil {
+			test.checkError = require.NoError
+		}
```
zmb3, line 109: Personally, I'd make sure to set `checkError` in all test cases above and then do away with this if. Less conditionals are easier to reason about and it makes your test cases more explicit.

tools/ami-cleanup/src/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,677 @@
...
+
+		t.Run(test.desc, func(t *testing.T) {
+			// Setup the application instance
```
zmb3, line 112:
```suggestion
			// Set up the application instance
```

tools/ami-cleanup/src/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,677 @@
...
+		{
+			"result 1",
+			"result 2",
+		},
+		{
+			"result 3",
+			"result 4",
+		},
+	}
+	allResults := make([]string, 0)
```
zmb3, line 142:
```suggestion
	allResults := make([]string, 0, len(actionSuppliedResults))
```

zmb3 (approved):
Thanks for the updates. This is looking a lot more like an idiomatic Go project now! 😄

PR #237, opened 2024-03-28
Added Renovate config for AMI cleanup tool

fheinecke:
Buddy PR for https://github.com/gravitational/shared-workflows/pull/214 to reduce the PR size. This PR adds Renovate to this repo to handle dependency updates for the AMI cleanup tool. The majority of these changes have been pulled from existing, working files in the cloud-terraform repo.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,0 +1,54 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:recommended",
+    ":enableRenovate",
+    ":semanticCommits",
+    ":enableVulnerabilityAlerts",
+    ":ignoreUnstable",
+    ":label(dependency-update)",
+    ":prConcurrentLimitNone",
+    ":prHourlyLimitNone",
+    ":prImmediately",
+    ":rebaseStalePrs",
+    ":renovatePrefix",
+    "helpers:pinGitHubActionDigests", // This ensures that underlying tags are not replaced with a separate commit (tags are immutable but commits are not)
+    "preview:dockerVersions",
+    "regexManagers:githubActionsVersions", // See https://docs.renovatebot.com/presets-regexManagers/#regexmanagersgithubactionsversions for how to use this
+    "github>gravitational/shared-workflows//.github/renovate/commitMessage.json5",
+    "github>gravitational/shared-workflows//.github/renovate/labels.json5",
+    "github>gravitational/shared-workflows//.github/renovate/languages.json5",
+    "github>gravitational/shared-workflows//.github/renovate/renovate.json5",
+    // Presets for each project
+    "github>gravitational/shared-workflows//tools/ami-cleanup/renovate.json5"
+  ],
+  "useBaseBranchConfig": "merge", // This is set to allow for some degree of testing PRs, see https://github.com/renovatebot/renovate/discussions/16108
+  // This is used so that not everything in the entire repo is setup for Renovate at once
+  // Initially this configuration is designed to only support the new GHA EKS clusters
+  "enabledManagers": [
+    "github-actions",
+    "custom.regex",
+    "gomod"
+  ],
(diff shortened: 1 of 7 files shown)


.github/renovate.json5
```diff
@@ -0,0 +1,54 @@
...
+    "github>gravitational/shared-workflows//.github/renovate/languages.json5",
+    "github>gravitational/shared-workflows//.github/renovate/renovate.json5",
+    // Presets for each project
+    "github>gravitational/shared-workflows//tools/ami-cleanup/renovate.json5"
+  ],
+  "useBaseBranchConfig": "merge", // This is set to allow for some degree of testing PRs, see https://github.com/renovatebot/renovate/discussions/16108
+  // This is used so that not everything in the entire repo is setup for Renovate at once
+  // Initially this configuration is designed to only support the new GHA EKS clusters
+  "enabledManagers": [
+    "github-actions",
```
jentfoo, line 29: Dependabot is already handling github-actions: https://github.com/gravitational/shared-workflows/blob/main/.github/dependabot.yml#L3

Why not just add to the Dependabot configuration for ecosystems which it's already managing?

fheinecke: See https://github.com/gravitational/shared-workflows/blob/main/rfd/0001-tooling-requirements.md#dependency-management

wadells: @jentfoo: We've generally found Dependabot lacking for internal tooling IaC needs.  It doesn't support critical ecosystems (helm, docker -- yes we've opened the feature requests) and lacks the configurability of renovate (e.g. in how we determine update sources).  I'm in favor of incrementally replacing dependabot with renovate in this repo.

jentfoo: I agree in the use of Renovate, I recognize it covers gaps that Dependabot has.  But it seems like for a given repo we should be using a single tool for a given ecosystem.  Having a single ecosystem (github actions) managed by two tools may add confusion.

fheinecke: I can replace dependabot with Renovate in this repo if you like. I'd be a bit more work but not a significant amount, given that the (current) one other project is pretty much just Go + GHA workflows.

jentfoo: It's my preference to keep them uniform for a single ecosystem + repo combo, unless there is a reason I am not considering to use both.

wadells: AFAIK we haven't messed with renovate for golang yet, so I'd recommend we keep dependabot for go. I'm fine switching all of GHA to renovate in this repo.

.github/renovate.json5
```diff
@@ -0,0 +1,54 @@
...
+    // These predate RFD 0001 and are managed by dependabot
+    "bot/**", 
+    ".github/workflows/codeql.yml",
+    ".github/workflows/csv-lint.yaml",
+    ".github/workflows/dependency-review.yaml",
+    ".github/workflows/github-action-lint.yaml",
+    ".github/workflows/govulncheck.yaml",
+    ".github/workflows/json-lint.yaml",
+    ".github/workflows/terraform-lint.yaml",
+    ".github/workflows/trivy.yaml"
```
wadells, line 44: Lets switch these to renovate.

.github/renovate.json5
```diff
@@ -0,0 +1,54 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:recommended",
```
wadells, line 4:
```suggestion
    "config:best-practices",
```

This adds the following on top of `config:recommended`:

```
{
  "configMigration": true,
  "extends": [
    "config:recommended",
    "docker:pinDigests",
    "helpers:pinGitHubActionDigests",
    ":pinDevDependencies"
  ]
}
```

https://docs.renovatebot.com/presets-config/

wadells (approved):
No concerns with getting this in and then iterating.

PR #239, opened 2024-03-28
Added CI/CD pipelines for AMI cleanup tool

fheinecke:
This is a buddy PR for https://github.com/gravitational/shared-workflows/pull/214 to reduce the PR size. This PR adds GHA PR testing workflows, and GHA release workflows that are backed by Earthfile targets.

This tool uses [Earthly](https://earthly.dev) for most of the build, review, and release process. We've only used this on a very limited basis for other small projects. I'm using it here (and possibly on future shared-workflow projects) to evaluate it as a _possible_ long-term solution to some of the build issues we've had with Teleport. If it ends up not working well then I can remove it/replace it with our more traditional makefiles/dockerfiles/shell script approach.

### DIFF
--- a/.github/workflows/ami-cleanup-cd.yaml
+++ b/.github/workflows/ami-cleanup-cd.yaml
@@ -1,0 +1,42 @@
+name: AMI cleanup tool CD
+on:
+  push:
+    tags:
+      - "ami-cleanup-v[0-9]+.[0-9]+.[0-9]+**"
+
+concurrency:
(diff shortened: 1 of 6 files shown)


r0mant (changes requested):
Let's please not introduce the new tool in the stack that everyone has to learn until we've gone through the proper proposal/design/review process and understood all the pros and cons and what it buys us (which is out of scope of this work IMO).

fheinecke:
@r0mant this isn't exactly new - we're using it here: https://github.com/gravitational/gha-exporter/blob/main/Earthfile

PR #255, opened 2024-05-17
Add tool to check PRs for vulnerabilities

fheinecke:
This PR adds a tool that will be run on all teleport/teleport.e PRs to perform static code analysis, looking for vulnerabilities. Right now it only supports Go via `govulncheck`, but it could be easily extended to support other languages as well.

The general process will look like this:
1. A PR is opened.
2. A workflow will call an action (not written yet) that runs this tool.
3. The tool will look to see if relevant files have changed (go.mod, go.sum, *.go).
4. If so, the tool will clone the repo. The tool has an argument that can be used to clone other repos, so PRs to teleport.e would clone Teleport, as well as it's submodules.
5. The tool identifies all Go modules associated with the changes. For example, for teleport.e product changes it would identify the go.mod file in the cloned teleport repo. For changes to tooling (which don't share a go.mod file with product changes), the tool would identify the associated go.mod file for the tooling change.
6. `govulncheck` is ran. If a vulnerability is found, then it the tool exits with a non-zero exit code, causing the action to fail. Note that this is ran against the entire module. We might want to only run it against specific packages that are changed.

There are several remaining things that I will file as separate PRs:
* Tests
* Makefile/Dockerfiles to build and release this tool
* Dependency update management for this tool
* Changelog
* GitHub Action that executes this tool
* Workflows in teleport/teleport.e that call the action for this tool

I have not tested this tool thoroughly, and will likely need to make minor revisions in an upcoming unit test PR.

Please note that most of this PR by line count is a combination of `go.sum` contents, and the LICENSE file.


r0mant (commented):
@fheinecke Why not just execute `govulncheck` in CI directly? This approach that imports it as a library means there's another tool we need to maintain and feels like a little too over-complicated for what we need for now.

fheinecke:
@r0mant please see [this Slack conversation](https://gravitational.slack.com/archives/C0351AUPJ8K/p1715914213203359?thread_ts=1715905682.247979&cid=C0351AUPJ8K).

In short, it boils down to a tradeoff between these three things:
* Cost
* Run time per PR
* Complexity (additional code, etc.)

Just running `govulncheck` on every PR would either be expensive, or take a long time. This PR minimizes this by adding a tool that only runs on modules (or alternatively packages with a minimal change) that are related to the PR.

fheinecke:
Business requirements have changed, so I'm closing this PR.

PR #264, opened 2024-07-01
Introducing changelog generation tool

doggydogworld:
This is a rewrite of the `changelog.sh` script to Go. This should make it easier to maintain and to align with the companies preference to write tooling in Go. This PR also includes the following:

* Improvements to handling of Enterprise changelogs
  * No longer prints `Enterprise:` if there are no Enterprise changelogs.
  * No longer generates a summary if `changelog` is not present in PR body.
  * No longer includes links to pull requests since those will be private.

Some notes for implementation:
* Introduced a small `gh` package to wrap the calls to the `gh` CLI. When looking to implement using the GitHub Web APIs directly I noticed that it would be much simpler with GitHub v4 API than with v3 and since that didn't really fit with our existing v3 library and would over-complicate the PR I figured better to just keep the simple implementation for now.
* Introduced a `git` package for some encapsulation around `git`


bot/tools/changelog/main.go
```diff
@@ -0,0 +1,220 @@
...
+
+	err := cmd.Run()
+
+	if err != nil {
+		return strings.TrimSpace(stderr.String()), trace.Wrap(err, "can't get last released version")
+	}
+	out := strings.TrimSpace(stdout.String())
+
+	return "v" + out, nil
+}
```
fheinecke, line 201: IMO it would be better to pass this in as a parameter rather than invoking `make`. This would allow the tool to be used in other projects without this target.

bot/go.mod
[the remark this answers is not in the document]
doggydogworld: Oh oops. My eyes are pretty bad so I saw the structure as something completely different. Thought it was odd to have `tools` nested under `bot`. I'll fix it up.

bot/tools/changelog/main.go
doggydogworld: So this is actually passed in as a flag (`--base-tag`). This is just the fallback if that flag isn't passed in. I tried to keep the functionality close to the current implementation. But yeah this would fail if the `--base-tag` flag isn't set and the Makefile doesn't exist or have that target making it not so portable.

For now to be clearer e.g. `--base-tag was not set, defaulted to invoking 'make version' but failed`. Should be less ambiguous and indicate a way to "fix" it.

go.mod
doggydogworld, line 1: @fheinecke I added the `go.mod` at the top of the directory to get my code in a working state. I wasn't quite sure what the intention was for the Go module. I did see in the file structure that there were some `go.mod` files in nested directories. To me that seemed kind of odd because I'd imagine that most of the code would be versioned together (or at least `libs` and `tools`). With `go.mod` setup in different directories we would have to use Go workspaces to manage the dependency between modules.

Oh also the `go.sum` is currently messed up a bit I'll have to fix it.

libs/gh/pull_request.go
```diff
@@ -0,0 +1,66 @@
...
+	FromDate string
+	ToDate   string
+}
+
+// ListChangelogPullRequests will search for pull requests that provide changelog information.
+func ListChangelogPullRequests(ctx context.Context, dir string, opts *ListChangelogPullRequestsOpts) ([]ChangelogPR, error) {
+	var prs []ChangelogPR
+	query := fmt.Sprintf("base:%s merged:%s -label:no-changelog", opts.Branch, dateRangeFormat(opts.FromDate, opts.ToDate))
+
+	data, err := RunCmd(dir, "pr", "list", "--search", query, "--limit", "200", "--json", "number,url,title,body")
```
r0mant, line 47: Since we're moving this to shared-workflows, we should use github package we use in other tools here rather than calling CLI: https://github.com/gravitational/shared-workflows/blob/main/bot/internal/github/github.go

libs/gh/gh.go
r0mant, line 1: Like mentioned below, let's prefer to use Github SDK rather than CLI. Then this file will not be needed.

tools/changelog/changelog.go
```diff
@@ -0,0 +1,144 @@
...
+
+const (
+	ossCLTemplate = `
+{{- range . -}}
+* {{.Summary}} [#{{.Number}}]({{.URL}})
+{{ end -}}
+`
+	entCLTemplate = `
+{{- range . -}}
+* {{.Summary}}
```
r0mant, line 52: Why does the enterprise template not include PR number/URL? External users can't access those PRs but I think we should still include it, it's useful for us.

tools/changelog/changelog.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return buff.String(), nil
+}
+
+// convertPRToChangelog will convert the list of PRs to a nicer format.
+func newChangelogInfoFromPR(pr gh.ChangelogPR) changelogInfo {
+	found, clSummary := findChangelog(pr.Body)
+	if !found {
+		// Pull out title and indicate no changelog found
+		clSummary = fmt.Sprintf("NOCL: %s", pr.Title)
```
r0mant, line 117: Do we need this? If the PR has no changelog, why include it? You always delete those anyway when creating a release (I do, anyways).

tools/changelog/main.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/gravitational/trace"
+
+	"github.com/gravitational/shared-workflows/libs/gh"
+	"github.com/gravitational/shared-workflows/libs/git"
+)
+
+var (
+	baseBranch = kingpin.Flag(
+		"base-branch",
+		"The base release branch to generate the changelog for.  It will be of the form branch/v*",
```
r0mant, line 38:
```suggestion
		"The base release branch to generate the changelog for. It will be of the form branch/v*",
```

tools/changelog/main.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	if err := gh.IsAvailable(); err != nil {
+		return trace.Wrap(err)
+	}
+	return nil
+}
+
+// getBranch will return branch if parsed otherwise will attempt to find it
+// Branch should be in the format "branch/v*"
+func getBranch(branch, dir string) (string, error) {
```
r0mant, line 119: Can these things (get branch, get forked branch, etc.) be moved into a shared reusable "git" library package, similar to what we have for "github"?

fheinecke: At that point it might just be easier to use https://github.com/go-git/go-git and remove the CLI tool dep entirely at the same time. I've used that library on a handful of projects before and it works well

r0mant: Yep, agreed.

doggydogworld: Yup I can do that should be a quick change.

tools/changelog/changelog.go
r0mant: Ah I think I misunderstood, this indicates the PR just doesn't have "changelog" line in the description, correct? If so,  that should be fine for now.

As a next step, we should enforce "changelog:" line or no-changelog label on all PRs going forward, we've wanted to do this for a while but haven't had time. But we can do this as a follow up.

doggydogworld: Yeah this is just a "warning" given to notify if there is a missing `changelog`. This isn't expected and I've only seen it occur in the Enterprise repo. I believe that the enforcement of the `changelog:` or `no-changelog` label isn't enabled there yet.

This just matches the current behavior but it could also be updated to spit out a warning to stderr in the future.

tools/changelog/changelog.go
doggydogworld: I removed it as it was indicated in the issue to remove it since it wasn't useful to external users. I had noticed that these links are also not present in the some of the newer Enterprise entries in the changelog.

libs/gh/gh.go
doggydogworld: Yeah I was just staging the packages here because I was a bit confused on the layout. Wasn't sure if it was intended to be multi-module or not. Will reimplement this in the `github` package using the SDK.

libs/gh/pull_request.go
doggydogworld: Something to mention is that packages cannot import `internal` unless they're nested in the same parent directory as `internal`. I'd imagine that everything under the `internal` package would be moved up under `libs`.

fheinecke: Yea it'd need to move to a new module under `/libs`.

tools/changelog/main.go
jimbishopp: fwiw - Be sure to check the go-git [compatibility guide](https://github.com/go-git/go-git/blob/master/COMPATIBILITY.md). I had to go with the [git lib](https://github.com/gravitational/cloud/blob/master/tcctl/release/git/git.go) route for a different project due to a lack of support for certain commands.

doggydogworld:
@r0mant @fheinecke @camscale 

Updated based on feedback:
* New `libs/git` module added
  * Wraps `go-git` library and provides method for shelling out to a single git repository (directory).
* New `libs/github` module added
  * Includes a new function that uses the v4 GitHub API to search for commits containing changelogs.
  * Temporary `z_github.go` file included to prepare for a future refactor of the bots internal library.
* Updated `tools/changelog`
  * Now requires a path to the local Teleport repository as an argument. This is to remove code that probably wasn't necessary.
  * Uses new `libs/git` and `libs/github` modules

tools/changelog/changelog.go
r0mant: I think we should keep including them.

tools/changelog/changelog.go
```diff
@@ -0,0 +1,142 @@
...
+	isEnt    bool
+	repo     string
+	ghclient *github.Client
+}
+
+// generateChangelog will pull a PRs from branch between two points in time and generate a changelog from them.
+func (c *changelogGenerator) generateChangelog(branch string, fromTime, toTime time.Time) (string, error) {
+	// Search github for changelog pull requests
+	prs, err := c.ghclient.ListChangelogPullRequests(
+		context.Background(),
```
r0mant, line 72: Nit: Pass context from the function caller.

libs/git/go.mod
r0mant, line 1: Why does every package have go.mod and go.sum?

libs/github/go.mod
```diff
@@ -0,0 +1,32 @@
+module github.com/gravitational/shared-workflows/libs/github
+
+go 1.22.4
+
+require (
+	github.com/cli/go-gh/v2 v2.9.0
+	github.com/google/go-github/v37 v37.0.0
+	github.com/gravitational/trace v1.3.1
+	github.com/shurcooL/githubv4 v0.0.0-20240429030203-be2daab69064
```
r0mant, line 9: Why do we need 3 different libraries for Github? Can we simplify? In our existing bot we already use github.com/google/go-github.

libs/git/git.go
```diff
@@ -0,0 +1,99 @@
...
+	return ref.Name().String(), nil
+}
+
+// GetParentReleaseBranch will attempt to find a parent branch for HEAD.
+// This will also work if HEAD has branched from an earlier commit in a release branch e.g.
+//
+//	 o---o---HEAD
+//	/---o---o---branch/v16
+func (r *Repo) GetParentReleaseBranch() (string, error) {
+	forkPointRef, err := r.RunCmd("merge-base", "--fork-point", "HEAD")
```
r0mant, line 87: It's kind of sad that we now have to use both git library and git binary. It feels like it's the worst of both worlds tbh because we still need to have git binary in path, but now also have an extra dependency.

Sorry to keep going back and forth on this, but should we revert back to just calling git CLI?

libs/git/go.mod
doggydogworld: It's in the RFD and was implemented per this [discussion](https://github.com/gravitational/shared-workflows/pull/264#discussion_r1661805402). I don't have enough context on the why but having separate Go modules would prevent API changes from causing dependents of a library to break.

libs/git/git.go
doggydogworld: No worries I did kind of have some misgivings on this. I was intending this to be as close to a 1-to-1 parity of the original `changelog.sh` as possible but it does end up with code that doesn't feel so clean. Unfortunately this part is not supported out of the box for the `go-git` library and might be possible to implement but would be complex.

I was thinking a better approach could be to have this be less opinionated on how this is obtained and just expect more from the arguments. In this case the git CLI just does a better job so we could just simplify the `changelog.sh` to do some of the trickier git stuff and pass it in as an argument to this tool.

I also did think maybe git is just too good at being git so maybe we could keep git commands in the bash script. This *does* much run faster though when it doesn't have to execute the git CLI.

libs/github/go.mod
doggydogworld: I was using the v4 API for the code but it isn't absolutely necessay so just moved it real quick to use the v3 API. The `go-gh` library has a useful function I just stumbled upon that will use look for `GITHUB_TOKEN`, github config file, or `gh auth`. Useful for running it on developer machines and CI.

I could remove that dependency. I need to update the config to note that `GITHUB_TOKEN` is needed anyway. Could have it maybe be a requirement to pass in the token.

libs/git/git.go
doggydogworld: Actually just thought of another solution. GitHub API does provide some methods for [querying git objects](https://docs.github.com/en/rest/git?apiVersion=2022-11-28). I think it should be enough for this use case. With that this changelog tool wouldn't need a dependency on `git` (and therefore the filesystem) at all. Any git features we do need can be passed in as arguments.

This would change the API of this changelog tool
```
changelog <base_branch> <base_tag>
```
Where `base_branch` and `base_tag` are now required arguments. Expected usage would still be `make changelog`.

libs/git/go.mod
r0mant: This is an overkill, let's not have a go.mod/sum per package.

r0mant (commented):
@doggydogworld We've been going back-and-forth on this for quite a bit now, I think we should be wrapping this up as it was supposed to be a just quick port from bash to Go. From my side I'd like us to address 3 things before it's good to go:

- Remove go.mod/sum from every package, this is very unnecessary.
- Let's go back to just calling `git` CLI and remove dependency on git library.
- Restore PR links in enterprise changes list (I left a comment about this before but it's still unaddressed).

Can you please make these changes and then let's get this ready to merge and start using.

go.mod
```diff
@@ -0,0 +1,27 @@
+module github.com/gravitational/shared-workflows
```
r0mant, line 1: I don't think this is correct still. We should have a go.mod per tool I feel, not a single one for the entire repo or one for each package.

go.mod
```diff
@@ -0,0 +1,27 @@
...
+
+go 1.22.4
+
+require (
+	github.com/alecthomas/kingpin/v2 v2.4.0
+	github.com/cli/go-gh/v2 v2.9.0
+	github.com/google/go-github/v37 v37.0.0
+	github.com/gravitational/trace v1.4.0
+	github.com/stretchr/testify v1.9.0
+	golang.org/x/oauth2 v0.21.0
```
r0mant, line 11: This looks much better as dependency list 👍

tools/changelog/changelog.go
```diff
@@ -0,0 +1,136 @@
...
+{{ end -}}
+`
+)
+
+var (
+	// clPattern will match a changelog format with the summary as a subgroup.
+	// e.g. will match a line "changelog: this is a changelog" with subgroup "this is a changelog".
+	clPattern = regexp.MustCompile(`[Cc]hangelog: +(.*)`)
+
+	clParsedTmpl = template.Must(template.New("oss cl").Parse(clTemplate))
```
r0mant, line 54:
```suggestion
	clParsedTmpl = template.Must(template.New("cl").Parse(clTemplate))
```

r0mant (commented):
Last comment about go.mod, then lgtm.

Also, please don't forget to update `make changelog` target in teleport repo to use this tool as a follow up.

go.mod
doggydogworld: Oh I think I have a better idea of what is wanted here then. I'm thinking then it should be a module for libs as a whole and a module for each tool.

For example in this change two modules will be created.
```
github.com/gravitational/shared-workflows/libs
github.com/gravitational/shared-workflows/tools/changelog
```

Tools are versioned separately. Everything under `libs` will be versioned together.

doggydogworld:
> Last comment about go.mod, then lgtm.
> 
> Also, please don't forget to update `make changelog` target in teleport repo to use this tool as a follow up.

FYI created a PR for this just to be ready [here](https://github.com/gravitational/teleport/pull/44099)

go.mod
doggydogworld: Btw I implemented the above. Check it out and let me know if it seems right.

r0mant (approved):

fheinecke:
@r0mant I disagree with using a single Go module under `libs/`. Using a single module _adds_ complexity, rather than reducing it, due to the additional coupling between libraries. This makes it more difficult for tools that depend on this "library module" to update, and removes our ability to defer updates in some cases. 

### Example 1:

We currently have separate "git" and "github" libraries inside the library module. This changelog generation tool depends on both of these libraries (regardless whether they're one or two modules). It is reasonable to expect that at some point we will introduce at least one breaking change to one of these libraries, for example, changing the signature of a function in the github library. It is also reasonable to expect that we may eventually discover a security issue with one of the libraries, for example, a CLI injection vulnerability in the git library.

When a breaking change is introduced, we may to avoid updating the changelog generation tool for awhile if the breaking change ends up being difficult to handle without providing a lot of value for this tool. If we defer this update, but we later discover a security issue with one of the other libraries, then we will be forced to take on both the security fix work and breaking change migration work at the same time, for all tools affected by both issues. If we use a separate module for each separate library, then we can avoid this extra work.

### Example 2:

Say we want to add a new feature to this tool, such as pushing these release notes to S3. We add a new library to wrap S3 API calls, which may be needed for things like handling pagination, retries, etc. If there is a breaking change in elsewhere in the library module, then migrating to accommodate this breaking change will now be required to add the new feature. This adds extra work where there otherwise wouldn't be any.

### Example 3 (experienced with Teleport dependencies):

Right now the library module is relatively simple with few dependencies, keeping it small. However, if the module adds a significant number of dependencies (see the above discussion about go-git), _and_ it adds one of the Go reflection footguns  (directly or indirectly via a dependency), then the binary size for every tool will explode. Using separate modules negates this issue as only dependencies (and dependencies of dependencies, and so on) that are needed by a given tool will be included.

At absolute bare minimum, this Go library module needs to move to a subdirectory of `libs`. The `libs` directory is not exclusively intended for Go code, and we should keep Go code/toolchain files out of it.

PR #266, opened 2024-07-16
Added local dev and release tooling to changelog tool

fheinecke:
This adds the following to the changelog tool:
* CI workflow that lints the project and runs all tests. This should be set as a required merge check after this PR lands.
* A CD workflow that builds tarballs and container images, and pushes to GH (release and GHCR).
* A makefile that can lint, test, build, and clean this project.

The intent with these files is that they should be easy to either copy/paste to new projects, or could easily be moved under `libs`. This is part of the up-front cost that I talked about in [the RFD for tooling requirements](https://github.com/gravitational/shared-workflows/blob/main/rfd/0001-tooling-requirements.md).

This does not add changelog generation, as that will likely depend on this tool. This project is also missing Renovate config, which I'll add in a future PR.

I'll target this at `main` once the current base branch lands.


fheinecke:
I _think_ the 403 response that is causing the build to fail is because [a repo-level setting needs to be changed](https://hackmd.io/@maelvls/fixing-403-forbidden-ghcr). I'll ask Roman or another admin to work with me on this later this week.

zmb3:
This feels like an awful lot of complexity for what was originally a simple script. Do we really need to build multi-arch images and store artifacts for multiple operating systems? Can we instead just tag releases and then `go run github.com/gravitational/shared-workflows/changelog@v0.1.2` this tool wherever we need it?

fheinecke:
> Do we really need to build multi-arch images and store artifacts for multiple operating systems?

Yes, especially because this will run on dev machines, local (macos arm64) and sometimes remote (linux arm64 and amd64). In addition, there isn't added complexity by building for multiple platforms if vs one, so IMO it'd be better to take care of it now than wait for it to become a problem that requires additional work.

This also makes the tooling easier to port to other projects (there are several that we need to migrate to this repo), which means that it'll be easier to manage multiple projects in the near future.

> Can we instead just tag releases and then go run github.com/gravitational/shared-workflows/changelog@v0.1.2 this tool wherever we need it?

IMO it'd be better to build the tool once and run the same binary (for a given platform) everywhere. This should help avoid the "it works/doesn' work on my machine" issue when the build is "tainted" by each user's environment. For example, devs across the org likely have several different versions of Go installed (some latest, some matching Teleport go.mod, some assorted versions). My understanding is that `go run` will compile the binary with whatever Go version the `go` binary is for, rather than what's specified in `go.mod` (please correct me if I'm wrong). This means that different devs will have different compiled code for the same source code tag, which can result in Go version-specific bugs, like we periodically have with Teleport when we change Go versions.

---

A large part of the RFD that this work is based off of is that it's better for us, as a company, to invest some up front time/effort into making it easier to build and maintain new tooling. That's essentially what this PR is - we do this now, and then when we add the next project (new or migrated), we can avoid this work. We'll simply copy/paste, or soft link, or "include" (via various file-specific mechanisms) these changes in projects going forward, and we'll have a complete set of local dev and CI/CD tools for these projects. It'll also be standardized, which will make it easier to understand how things work, lowering the cost of making changes in the future.

zmb3:
> Yes, especially because this will run on dev machines, local (macos arm64) and sometimes remote (linux arm64 and amd64).

`go run` will just work on any machine that has Go installed. No need to store images, purge old ones because storage costs get too high, etc.

> when the build is "tainted" by each user's environment

If you're concerned about the user's environment (and not about Go behaving differently with two invocations against the same source code) then we can `go run` inside a docker image to isolate the user's environment.

> This also makes the tooling easier to port to other projects

I've never been a big fan of the argument that adding complexity now helps support doing more complicated things in the future. I think we should be doing the opposite - fighting against complexity and striving to do the simplest thing possible every chance we get.

> This should help avoid the "it works/doesn' work on my machine" issue 

This seems like a theoretical argument. Do you have any practical concerns about a Go program as simple as this one not working consistently across environments?

> My understanding is that go run will compile the binary with whatever Go version the go binary is for, rather than what's specified in go.mod (please correct me if I'm wrong)

In the past you would have been correct, but as of Go 1.21 this is no longer true. The `go` tool will use whatever version of the toolchain is specified in go.mod (downloading the correct version in the background if necessary).

fheinecke:
> No need to store images, purge old ones because storage costs get too high, etc.

When building locally for darwin/arm64 the image size is 17MB:
```console
fredheinecke@Freds-MacBook-Pro changelog % docker image ls changelog
REPOSITORY   TAG       IMAGE ID       CREATED         SIZE
changelog    latest    ef3cd8e05d5f   7 seconds ago   17MB
```

This isn't terribly large, and is dwarfed by other more frequently updated images that devs typically have on their machines (such as Teleport buildbox). Additionally, we have historically taken the approach of "throw money at problems until they go away" with things like storage space in cloud/SaaS services, so I don't think that the added GHCR cost is significant here.

> If you're concerned about the user's environment (and not about Go behaving differently with two invocations against the same source code) then we can go run inside a docker image to isolate the user's environment.

Maybe I'm missing something, but doesn't this just shift the build process towards the users of this tool (our devs)? We would still need the vast majority of the logic (such as dockerfile and build scripts), but now they'd both need to exist locally and be executed when users run this tool.

> fighting against complexity and striving to do the simplest thing possible every chance we get

This does fight complexity. We've rewritten this PR's tooling several times now across several projects (in fact, the majority of this PR is copy/pasted from our other repos with minor changes). Every time I go to make a change to any of these tools, I have to relearn/rediscover how they're built and how the tooling for them works. Given how many repos I touch (1-2 dozen per month based on my PR history over the past few months), and how many tools I touch (in our larger repos like Teleport there are many in each), this is really painful for me. Using one standard build process for all the relatively simple projects that I work on makes the maintenance for each one much much easier.

I can provide some specific examples projects in repos across our org we want this general build outcome (binaries and/or container images) but do so in very different ways, if desired.

> Do you have any practical concerns about a Go program as simple as this one not working consistently across environments?

I don't have any explicit examples. The Go version one (which you corrected me on - thanks for the information) was the only specific one off the top of my head. However, as with most software, specific bugs are difficult to predict but bugs in general always occur. Building once and using the build everywhere removes most risk of user-specific bugs.

doggydogworld:
Also fwiw while this might be overkill for the usage of the changelog tool I think that having a standardized template for all workflows is important. Tools in the future might be expected to run in a container or on different distributions and to have a workflow that has already been tested on a simple tool like this will make it easier to handle more complicated tooling.

IMO the only wrong way to deliver artifacts is to limit what we deliver.

doggydogworld (approved):

camscale:
There is a simple alternative to this which I've mentioned to the team before: hermit https://github.com/cashapp/hermit

Set up the repo once for hermit:

    hermit init --no-git

Activate the hermit environment (often done automatically if set up in your shell rc):

    source bin/activate-hermit

Install specific versions of tools:

    hermit install gitleaks@8.18.4 lefthook@1.7.2
    git add bin && git commit -m "hermit: Add gitleaks and lefthook tools"

The tools are downloaded on demand when run, the version is pinned to the repo and are easily updated.

This gives us pinned versions, automatic installation and no need to manage our own builds.

Just putting it out there as an idea.

(I use hermit locally for per-repo versions of tools, which I've needed when doing releases - different versions of rust, different versions of helm-unittest, different versions of node, etc. I do this by putting `*` in `bin/.gitignore` and `.hermit/.gitignore`, as hermit creates `bin` and `.hermit`, but I don't want to have a dirty repo for not checking in my local tooling).

.github/workflows/changelog-cd.yaml
```diff
@@ -0,0 +1,134 @@
...
+name: Changelog generator CD
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - tools/changelog
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+**-changelog"
```
camscale, line 11: when using multiple Go modules in a repository, the tag format should be <subdirs>/<version>: https://go.dev/wiki/Modules#faqs--multi-module-repositories

So in this case, this should be `"tool/changelog/v[0-9]+.[0-9]+.[0-9]+"`

.github/workflows/changelog-cd.yaml
```diff
@@ -0,0 +1,134 @@
...
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - name: Create event-specific values
+        id: setup
+        run: |
+          # Determine if the workflow was triggered via a push to main or a tag
+          # and get the version based off of that
+          if [[ "${GITHUB_REF}" =~ refs/tags/.* ]]; then
+            # Transforms tag refs like refs/tags/v1.2.3-changelog into v1.2.3
+            TAG="${GITHUB_REF#refs/tags/}"
+            echo "version=${TAG%-changelog}" >> "${GITHUB_OUTPUT}"
```
camscale, line 38: this will need to change too WRT my comment on multi-module tagging

tools/changelog/Makefile
```diff
@@ -0,0 +1,32 @@
+OS ?= $(shell uname -s | tr '[[:upper:]]' '[[:lower:]]')
+ARCH ?= $(shell uname -m)
+TOOL_NAME ?= changelog
+BUILD_DIR ?= build/$(OS)/$(ARCH)/
+VERSION ?= v0.0.0-dev
```
camscale, line 5: It is not typical to use `?=` to assign variables as it allows environment variables to override make variables, and this can cause unusual build issues as different people may have different environments.

You can define these with `=` and they are still overridable on the `make` command line. e.g:
```
make tarball ARCH=amd64 OS=linux
```

This still overrides these default values as command line values override makefile variables.

tools/changelog/Makefile
```diff
@@ -0,0 +1,32 @@
...
+
+container-image: OS = linux
+container-image: binary
+	@docker buildx build --platform="linux/$(ARCH)" -t "$(TOOL_NAME)" .
+
+clean:
+	@rm -rf build/
+	@if [ -n "$(shell docker image ls --quiet --filter "reference=$(TOOL_NAME):$(VERSION)" )" ]; then \
+		docker image rm -f "$(TOOL_NAME):$(VERSION)"; \
+	fi
```
camscale, line 30: There's no need to check that the image exists - `-f` prevents it from failing like it does for rm. I'd just do the removal without the test.

tools/changelog/Makefile
```diff
@@ -0,0 +1,32 @@
...
+
+test:
+	@gotestsum --format github-actions ./... -- -shuffle on -timeout 2m -race
+
+binary:
+	@echo "Building for $(OS)/$(ARCH) and writing to $(BUILD_DIR)"
+	@mkdir -p "$(BUILD_DIR)"
+	@GOOS=$(OS) GOARCH=$(ARCH) go build -o "$(BUILD_DIR)/" -ldflags="-s -w"
+
+tarball: TARBALL_NAME = "$(TOOL_NAME)-$(VERSION)-$(OS)-$(ARCH).tar.gz"
```
camscale, line 18: Don't use quotes when setting make variables - they are not special and just end up literally in the value. They will be placed on the command line literally and you will end up unquoted as it is expanded inside quotes already (the quotes in the var will close the quotes in the command).

camscale (commented):
Not sure where the discussion for this goes now, but here's my review FWIW

.github/workflows/changelog-cd.yaml
```diff
@@ -0,0 +1,135 @@
...
+          # and get the version based off of that
+          if [[ "${GITHUB_REF}" =~ refs/tags/.* ]]; then
+            # Transforms tag refs like refs/tags/tools/changelog/v1.2.3 into v1.2.3
+            echo "version=${GITHUB_REF#refs/tags/tools/changelog}" >> "${GITHUB_OUTPUT}"
+
+            # Eventually the parse-version action from the teleport.e repo
+            # should move into this repo and replace this logic
+            echo "should-release=true" >> "${GITHUB_OUTPUT}"
+            
+            if [[ "${GITHUB_REF%-*}" == "${GITHUB_REF}" ]]; then
```
camscale, line 44: This test looks backwards to me - if you strip off a `-*` suffix and it is the same as itself. that means it did not have the suffix and is not a prerelease.

This could more simply just match against a glob: `[[ "${GITHUB_REF}" == *-* ]]` (that is safe inside `[[` - it won't match files or anything unwanted like that)

.github/workflows/changelog-cd.yaml
```diff
@@ -0,0 +1,135 @@
...
+          fi
+
+          readarray -d '' RELEASE_TARBALLS < <(
+            find . -name '*.tar.gz' -print0
+          )
+
+          echo "Creating a release for ${VERSION} with files:"
+          for file in "${RELEASE_TARBALLS[@]}"; do
+            ls -lh "${file}"
+          done
```
camscale, line 131: No need for a `for` loop here. You can just run `ls -lh "${RELEASE_TARBALLS[@]}"`

fheinecke: Thanks - think I left that behind when I was working on the array building logic

camscale (approved):

r0mant (approved):

PR #271, opened 2024-09-05
Using correct attribute for PR URL

doggydogworld:
Was using the wrong attribute that represented the v3 REST API URL. Will now return the HTML URL which is the normal UI.


libs/github/pull_request.go
```diff
@@ -72,7 +72,7 @@ func (c *Client) ListChangelogPullRequests(ctx context.Context, org, repo string
 			Body:   pull.GetBody(),
 			Number: pull.GetNumber(),
 			Title:  pull.GetTitle(),
-			URL:    pull.GetURL(),
+			URL:    pull.GetHTMLURL(),
```
r0mant, line 75: Don't we have a test fixture for the changelog output that could've caught this issue?

r0mant (approved):

libs/github/pull_request.go
doggydogworld: The PR struct gets converted to a simpler structure to match what we had in the previous changelog tool. I had abstracted it since it because I wasn't sure if we would use v4 or v3 API (it originally used v4). Since it's using v3 API now I can add a test fixture for that so it should catch similar issues and maybe make it easier to see what data is available.

PR #273, opened 2024-09-24
Fixing the time format for parsing git timestamps

doggydogworld:
Previously would support timestamps with offsets but not timestamps in UTC time with "Z".

E.g. `2024-09-10T13:30:47-05:00` was supported but the equivalent in UTC time `2024-09-10T18:30:47Z` was not.

Also added some testing to prove it works and to provide a pattern for any future tests.


r0mant (approved):

doggydogworld:
Duplicate. Fixed in https://github.com/gravitational/shared-workflows/pull/274

PR #274, opened 2024-09-25
libs: Fix some git/github date/time issues

camscale:
Use the standard `time.RFC3339` time format with `time.Parse()` so that
we support dates with a `Z` suffix instead of just a UTC offset. The
simple `Z` format is allowed by ISO8601 format so we should parse that.
It turns out that the `%cI` format passed to `git show` will return the
format with the `Z` suffix.

This fixes an error when running the changelog tool:

    parsing time "2024-09-25T02:35:11Z" as "2006-01-02T15:04:05-07:00": cannot parse "Z" as "-07:00"

Change the form of the github query date formatting to use `>=` instead
of `>` when just a start date is provided. The makes it consistent with
the range format (`a..b`) used when an end date is provided as the range
includes the first date. The changelog tool will need to be updated to
handle this by bumping the start dates by 1 second, but this should fix
an issue where an Enterprise commit for a PR that gets a release tag
ends up listed in two releases.

---

Once merged, I'll tag this as `libs/v0.1.3` and raise a PR that updates
`tools/changelog` to work with the changed date range logic. This should
resolve an issue that @stevenGravy reported where an Enterprise
changelog entry appeared on the release after it went in, due to being
right on the boundary/tag of the v16.4.0 release.

### DIFF
--- a/libs/git/git.go
+++ b/libs/git/git.go
@@ -30,11 +30,6 @@ type Repo struct {
 	dir string
 }
 
-const (
-	// git should be expected to to output in strict ISO 8601 format
-	gitTimeFormat = "2006-01-02T15:04:05-07:00"
-)
-
 // NewRepoFromDirectory initializes [Repo] from a directory.
 func NewRepoFromDirectory(dir string) (*Repo, error) {
 	return &Repo{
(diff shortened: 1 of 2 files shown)


fheinecke (approved):

doggydogworld:
Oh I did part of the same fix yesterday haha: https://github.com/gravitational/shared-workflows/pull/273

I'll close mine out since yours has a bit more fixes. Had the fix on my local but was trying to finish out some goals so forgot to get the PR our earlier.

doggydogworld (approved):

camscale:
> I'll close mine out since yours has a bit more fixes.

You have tests though. I thought about adding tests but it was a 1 char change and 11pm. I'll take a copy of the tests from your PR and add them to this.

camscale:
I updated the first commit to get rid of the "custom" time format as it is the same as `time.RFC3339`, so I just use that instead. I started moving the tests from the original PR but realised all I was really testing was that `time.RFC3339` did what it said it did which seemed a little pointless, so I've left those tests behind.

r0mant (approved):

PR #275, opened 2024-09-30
CI/CD environment value loader

fheinecke:
This is a draft of a tool to assist with the removal/replacement of GHA environment secrets/values.

> [!IMPORTANT]
> Regardless of the outcome of this review, I will not be merging this PR directly. I am _only_ looking for a review of the general approach, not a line-by-line review of the tooling itself.

See https://github.com/gravitational/teleport.e/pull/5085 for an example implementation of this tool in one of our CI/CD pipelines.

This tool is designed to load context-specific secrets and values into **both** CI/CD pipeline environments, and local environments. This tool will take in:
* The name of an "environment", such as `local`, `staging`, or `production`
* A context-specific "value set", such as `build-linux` or `publish-os-packages`

With this information, it will find and load appropriate YAML key/value files that are stored the repo that this tool is invoked in. For secrets, the tool will attempt to load matched SOPS files, which should be encrypted using a KMS key. Access to this KMS key can be restricted to specific GHA environments and workflows, where we can add an approval flow to restrict access. Cloudtrail will record `Decrypt` events for the key, providing an audit log.

Here's the flow for adding or modifying values for existing workflows:
1. Create a new teleport/teleport.e/cloud/etc. branch
2. Request access to an AWS role with KMS key access (only required for secrets)[^1]
3. Modify the appropriate values YAML file
4. Make any other context-specific changes needed (such as using the new value in a workflow)
5. Commit and file a PR

The workflow for adding a completely new environment or "value set" secret is similar, except it will require an additional PR to cloud-terraform to deploy a new KMS key and IAM role (to limit secret access to a specific context), as well as a PR to github-terraform to set a GHA environment value for the IAM role.[^2] This process (new workflow and new workflow-specific secret) is infrequent.

This should be both a much simpler, and much more secure approach than using AWS Secrets Manager because:
* If we use AWS SM, we need to store the ARNs of the secrets somewhere, and add additional tooling to load and use this information. The ARNs would be a part of the chain of trust (though not sensitive values) causing a "chicken and egg" issue.
* By placing all the environment values in the appropriate repos, it will be much easier to add new values when iterating on pipelines. New values can be added in the same PR as actual workflow changes. The alternative with AWS SM would be:
  1. Create and merge a PR in cloud-terraform to deploy a new AWS SM secret resource, including adding KMS and IAM permissions where appropriate
  2. Request access to a role that has access to the created AWS SM secret resource
  3. Manually upload the value via AWS console or CLI
  4. Create a new teleport/teleport.e/cloud/etc. branch
  5. Make any other context-specific changes needed (such as using the new value in a workflow)
  7. Commit and file a PR
  8. Hope that the name/value doesn't have to change as a result of a PR review, if so, go back to (1)
* PR reviewers can check the secret value at review time. This isn't possible if secrets are manually uploaded somewhere without peer review (AWS SM).
* Environment values will now be specific to a given release branch, removing the need for the `LEGACY_BRANCH_v###_` type variables that we periodically need. Once a branch is "expired", we will no longer need to do manual env-specific cleanup.

[^1]: There is a _slightly_ more complex way to do this where read access is not required, if we want to take this route.
[^2]: This doesn't pose a security risk because if the value is replace by _any_ value other than the appropriate IAM role, then the secret will just fail to decrypt at runtime.


tools/env-loader/pkg/envloader.go
```diff
@@ -0,0 +1,126 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/gravitational/shared-workflows/tools/env-loader/pkg/loaders"
+	"github.com/gravitational/trace"
+)
+
+// This is the path, relative to the git repo root, where environment
+// directories can be found.
+const CiDirectoryRelativePath = "environments"
```
doggydogworld, line 13: nit: Would prefer `.environments` since I usually associate hidden files with config.

tools/env-loader/pkg/envloader.go
```diff
@@ -0,0 +1,126 @@
...
+
+func findGitRepoRoot() (string, error) {
```
doggydogworld, line 15: Not sure if there's much value in having the tool search for repo root since it could be passed in as an argument. If the tool was less reliant on the git repository structure it I think it would also be more flexible in its usage. (i.e. `environments` config in subfolders, config passed into containers)

I have seen this pattern in some of our other tooling though so maybe it's a common use case. In which case it would maybe be better somewhere in `libs`.

doggydogworld:
Overall I really like this approach. I think it's more ergonomic than our current process and removes needless complexity in the configuration management for our pipelines.

It should also help with reproducibility since I believe the resulting dotenv file can be sourced e.g.

```console
$ env-loader > .env
$ source .env
$ echo $SECRET
SECRET_VALUE
```

Don't know about the portability of the above but did a quick test in bash and it seems to work.

tools/env-loader/pkg/envloader.go
fheinecke: Maybe I could make this an arg, and default to git root if not provided? This might make it more flexible while providing a sane default. WDYT?

fheinecke:
> I believe the resulting dotenv file can be sourced

Yep - I implemented dotenv as the first "writer" because:
* It's a really easy/simple spec
* It can be sourced via bash
* It can be `include`d in makefiles (should we need this at some point)

camscale:
What do you think of having a single environment file instead of splitting it up over many files? Keeping many files in sync I think is going to be harder and more error prone:

```
environments:
  - local
  - stage
  - prod
values:
  ARTIFACT_SOURCE_BUCKET:
    stage: tp-603330915152-us-west-2-build-artifacts-stage
    prod: dronestorage.gravitational.io
    local: '???'
  ARTIFACT_UPLOAD_AWS_ROLE:
    stage: arn:aws:iam::603330915152:role/tf-build-artifacts-upload-gha-stage-teleport.e
    prod: arn:aws:iam::126027368216:role/tf-os-package-repo-tool-dronestorage-upload-gha
  GPG_SIGNING_ARCHIVE:
    stage: ENC[AES256_GCM,data:PFzX2Q8mBsBfEDNJlgbTePYQcFM=,iv:FFls0uZktGsD4ueL1FFxXn6PnbqUpyJ5IkKn0FBtwbM=,tag:vIxYtwh/Vf09Omg0yOGRlA==,type:str]
    prod: ENC[AES256_GCM,data:PFzX2Q8mBsBfEDNJlgbTePYQcFM=,iv:FFls0uZktGsD4ueL1FFxXn6PnbqUpyJ5IkKn0FBtwbM=,tag:vIxYtwh/Vf09Omg0yOGRlA==,type:str]
valuesets:
  build-linux:
    - ECR_PRIVATE_REPOS_REGION
    - ECR_PROD_PULL_ROLE_ARN
    - RELCLI_IMAGE
    - ARTIFACT_UPLOAD_AWS_ROLE
    - ARTIFACT_SOURCE_BUCKET
```

I know this does not fully encapsulate the matrix dimensions (there's a build vs publish distinction too) so perhaps it gets overly cumbersome with too many dimensions, but I find the above easier to understand over lots of little files spread out.

I'm not sure I understand the use of `local` as an environment - I'm not sure what the "local" value of any of these vars would be.

In the other PR, there are separate file for each workflow x environment that specifies the value sets. I'm not sure that is needed - these are meant to be parameters which should be the same sets of vars regardless of environment.

fheinecke:
That would certainly make it easier to manage. However it has several downsides:
* SOPS (which is the primary secrets source that we're currently planning on using) does not support separate keys for for different values. This means that if a single workflow with access to the SOPS key for the file is compromised, then all values are compromised.
* We can't control ownership in a Git repo very well, with all values in one file. For example, with this approach it's not possible to set us as codeowners for "releases" values, and senior leadership as codeowners for "PR reviewers" values.

I think the current approach is a compromised between one monolithic file and completely separate files (with duplicate values in many cases). The current approach supports repo-level and environment-level `common.*` files, which are always loaded (with the value set taking precedence over these). 95% of our values are "GHA variables" rather than "GHA secrets". The implication here is that we don't care about hiding these values. Because of this, for each environment most values can live in a `common.yaml` file, rather than be copy/pasted between multiple value set files.

> I'm not sure I understand the use of local as an environment - I'm not sure what the "local" value of any of these vars would be.

The idea is that this can (eventually) hold values that the typical dev use to run workflows locally. Right now there are several places where we have logic to handle this in our workflows, which this should reduce or remove as well as centralize.

For example, most devs won't have access to the KMS key used for encrypting the GPG signing key SOPS file in either production or staging. They likely don't (or shouldn't if they do) have access to push to production or staging S3 buckets. This environment could:
* Store "secrets" (that aren't really secrets given the scope of their sharing) in non-encrypted files. For example, the "local" GPG signing key could be stored in plain text, allowing for easier debugging and development in some cases.
* Provide values that developers can access. For example, instead of prod/staging S3 bucket, it could point to a developer S3 bucket in teleport-dev, or perhaps a S3 service (such as minio) or S3 mock service stood up by the makefile.

fheinecke:
Closing this as the actual PR for this work is open at https://github.com/gravitational/shared-workflows/pull/280

PR #278, opened 2024-10-09
Do not require changelog entry for docs

ptgott:
Docs-only changes do not require a changelog entry, as we reserve the
changelog for changes to the product. Edit the changelog workflow to
ignore docs-only changes.


ptgott:
Actually, might make sense to add file-checking logic to the changelog workflow instead of this indirect route.

zmb3:
> Actually, might make sense to add file-checking logic to the changelog workflow instead of using this indirect route.

+1

We control both the labeling workflow and the check for a changelog entry. 

I'd rather have the changelog check simply not care about changelogs for docs changes.

zmb3 (approved):

fheinecke (changes requested):
This needs changes to maintain compatibility with release tooling, as discussed on slack here: https://gravitational.slack.com/archives/C0351AUPJ8K/p1738617752346529

bot/internal/bot/label.go
```diff
@@ -66,6 +66,12 @@ func (b *Bot) labels(ctx context.Context, files []github.PullRequestFile) ([]str
 		labels = append(labels, string(prSize(files)))
 	}
 
+	c := classifyChanges(b.c, files)
+	if c.Docs && !c.Code {
+		log.Println("Label: Adding no-changelog because this is a docs-only change.")
+		labels = append(labels, "no-changelog")
```
fheinecke, line 72: Can you change the value to [this constant](https://github.com/gravitational/shared-workflows/blob/main/bot/internal/bot/changelog.go#L30)?

bot/internal/bot/label_test.go
```diff
@@ -56,7 +56,7 @@ func TestLabel(t *testing.T) {
 					Deletions: 10,
 				},
 			},
-			labels: []string{"documentation", string(small)},
+			labels: []string{"documentation", "no-changelog", string(small)},
```
fheinecke, line 59: Same as above (const vs magic string)

fheinecke (approved):
LGTM outside of magic string

bot/internal/bot/label_test.go
ptgott: Fixed in 9078af911d64ccf6459280851bf0fb783acc7258

bot/internal/bot/label.go
ptgott: Fixed in 9078af911d64ccf6459280851bf0fb783acc7258

PR #280, opened 2024-10-11
Add `env-loader` tool

fheinecke:
This is a first iteration of the `env-loader` tool, to be used to help replace github environments as a root of trust. I will add everything under `Future work` in the readme as separate PRs.

Not adding more detail to the PR body to help ensure that the README has sufficient documentation.


tools/env-loader/cmd/env-loader.go
```diff
@@ -0,0 +1,101 @@
...
+	"maps"
+	"slices"
+
+	"github.com/alecthomas/kingpin/v2"
+	envloader "github.com/gravitational/shared-workflows/tools/env-loader/pkg"
+	"github.com/gravitational/shared-workflows/tools/env-loader/pkg/writers"
+	"github.com/gravitational/trace"
+)
+
+const ENV_VAR_PREFIX = "ENV_LOADER_"
```
zmb3, line 31:
```suggestion
const EnvVarPrefix = "ENV_LOADER_"
```

Go generally doesn't use SCREAMING_SNAKE_CASE for constants like you might be used to with other languages.

tools/env-loader/cmd/env-loader.go
```diff
@@ -0,0 +1,101 @@
...
+
+type config struct {
+	Environment string
+	ValueSet    string
+	Values      []string
+	Writer      string
+}
+
+func parseCli() *config {
```
zmb3, line 40:
```suggestion
func parseCLI() *config {
```
https://go.dev/wiki/CodeReviewComments#initialisms

tools/env-loader/pkg/envloader.go
```diff
@@ -0,0 +1,158 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/gravitational/shared-workflows/tools/env-loader/pkg/loaders"
+	"github.com/gravitational/trace"
+)
+
+// This is the path, relative to the git repo root, where environment
+// directories can be found.
+const CiDirectoryRelativePath = ".environments"
```
zmb3, line 29:
```suggestion
const CIDirectoryRelativePath = ".environments"
```

tools/env-loader/pkg/envloader.go
```diff
@@ -0,0 +1,158 @@
...
+	environmentsPath := filepath.Join(repoRoot, CiDirectoryRelativePath)
+	if _, err := os.Stat(environmentsPath); err != nil {
+		return nil, trace.Wrap(err, "failed to find environments path at %q", environmentsPath)
+	}
+
+	return FindEnvironmentFilesInDirectory(environmentsPath, environment, valueSet)
+}
+
+func loadEnvironmentValuesFromPaths(valueFilePaths []string, err error) (map[string]string, error) {
+	if err != nil {
```
zmb3, line 122: Why do we pass an error in at all if it means we always return an error back out?

Can the caller rather not call this method if it already knows there was an error?

tools/env-loader/pkg/loaders/subloader.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (sl *SubLoader) GetEnvironmentValues(bytes []byte) (map[string]string, error) {
+	subloader := sl.getSubloader(bytes)
+	if subloader == nil {
+		return nil, trace.Errorf("found no YAML loaders for the provided content")
+	}
+
+	environmentValues, err := subloader.GetEnvironmentValues(bytes)
+	if err != nil {
+		return nil, trace.Wrap(err, "failed to unmarshal YAML bytes with %q YAML loader", subloader.Name())
```
zmb3, line 41: This error is a bit confusing to me as nothing about the surrounding code implies anything about a YAML loader. Am I missing something?

tools/env-loader/pkg/loaders/loader.go
```diff
@@ -0,0 +1,30 @@
...
+	// Returns true if the byte array can be converted into a key/value map
+	CanGetEnvironmentValues([]byte) bool
+	// Converts the byte array into a key/value map
+	GetEnvironmentValues([]byte) (map[string]string, error)
+	// Human-readable name of the loader
+	Name() string
+}
+
+// Generic loader that can load values of any supported format
+var DefaultLoader = NewSubLoader(NewYamlLoader())
```
zmb3, line 30:
```suggestion
var DefaultLoader = NewSubLoader(NewYAMLLoader())
```

tools/env-loader/pkg/loaders/yaml.go
```diff
@@ -0,0 +1,93 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package loaders
+
+import (
+	"github.com/getsops/sops/v3/decrypt"
+	sopsYaml "github.com/getsops/sops/v3/stores/yaml"
```
zmb3, line 21:
```suggestion
	sopsyaml "github.com/getsops/sops/v3/stores/yaml"
```

General idiom is to keep package names (even when aliased) all lower case.

tools/env-loader/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "{}"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2015 Gravitational, Inc.
```
zmb3, line 189: 2015?

fheinecke: Copied this from a file in the teleport repo - fixed, thanks

tools/env-loader/pkg/loaders/subloader.go
fheinecke: Good catch. I refactored this while developing and missed the error message. Removed the "YAML" references.

tools/env-loader/pkg/envloader.go
fheinecke: > Why do we pass an error in at all if it means we always return an error back out?

It's not always returned back out - it's only returned if the error is non-nil, and it's wrapped. I took this approach to reduce some repetition in where the function is called. I can either do this:

```go
func loadEnvironmentValuesFromPaths(valueFilePaths []string, err error) (map[string]string, error) {
        if err != nil {} // Handle error
        // do the rest
}

func LoadEnvironmentValues(environment, valueSet string) (map[string]string, error) {
	return loadEnvironmentValuesFromPaths(FindEnvironmentFiles(environment, valueSet))
}

func LoadEnvironmentValuesInDirectory(directory, environment, valueSet string) (map[string]string, error) {
	return loadEnvironmentValuesFromPaths(FindEnvironmentFilesInDirectory(directory, environment, valueSet))
}
```

or I can do this:
```go
func loadEnvironmentValuesFromPaths(valueFilePaths []string) (map[string]string, error) {
        // do the rest
}

func LoadEnvironmentValues(environment, valueSet string) (map[string]string, error) {
        filePaths, err := FindEnvironmentFiles(environment, valueSet)
        if err != nil {} // Handle error
	return loadEnvironmentValuesFromPaths(filePaths)
}

func LoadEnvironmentValuesInDirectory(directory, environment, valueSet string) (map[string]string, error) {
        filePaths, err := FindEnvironmentFilesInDirectory(directory, environment, valueSet)
        if err != nil {} // Handle error
	return loadEnvironmentValuesFromPaths(filePaths)
}
```

Alternatively I can pass an anonymous function arg and capture the `FindEnvironmentFiles*` args via a closure, but this adds more boilerplate.

I can take whichever path you'd prefer - just trying to avoid copy/pasting identical error handling code.

tools/env-loader/cmd/env-loader.go
```diff
@@ -0,0 +1,101 @@
...
+
+func run(c *config) error {
+	// Load in values
+	envValues, err := envloader.LoadEnvironmentValues(c.Environment, c.ValueSet)
+	if err != nil {
+		return trace.Wrap(err, "failed to load all environment values")
+	}
+
+	// Filter out values not requested
+	if len(c.Values) > 0 {
```
doggydogworld, line 76: This length check here is unnecessary.

tools/env-loader/cmd/env-loader.go
```diff
@@ -0,0 +1,101 @@
...
+		maps.DeleteFunc(envValues, func(key, _ string) bool {
+			return !slices.Contains(c.Values, key)
+		})
+	}
+
+	// Build the output string
+	writer := writers.AllWriters[c.Writer]
```
doggydogworld, line 83: Need to check if `writer` is nil after this.

tools/env-loader/pkg/writers/writer.go
```diff
@@ -0,0 +1,36 @@
...
+	// Human-readable name of the writer, usually the output format.
+	Name() string
+}
+
+var (
+	dotenvWriter  = NewDotenvWriter()
+	DefaultWriter = dotenvWriter
+
+	// A map of all writers available.
+	AllWriters = map[string]Writer{
```
doggydogworld, line 33: nit: Not sure if `AllWriters` is a good name here. Maybe `FromName` or something equivalent would be more descriptive. I just saw `writers.AllWriters` above and I had to look at the type signature to figure out what this does.

tools/env-loader/cmd/env-loader.go
```diff
@@ -0,0 +1,101 @@
...
+		Short('v').
+		Envar(EnvVarPrefix + "VALUE").
+		StringsVar(&c.Values)
+
+	kingpin.Flag("format", "Output format of the value(s)").
+		Short('f').
+		Envar(EnvVarPrefix+"FORMAT").
+		Default("dotenv").
+		EnumVar(&c.Writer, slices.Collect(maps.Keys(writers.AllWriters))...)
+
```
doggydogworld, line 64: I think there's a `kingpin.Parse()` missing here.

tools/env-loader/pkg/writers/writer.go
fheinecke: This shouldn't occur - kingpin CLI arg validation will ensure that the key exists in the map.

```console
$ go run ./cmd/env-loader.go -f ""
env-loader: error: enum value must be one of dotenv, got '', try --help
exit status 1
$ go run ./cmd/env-loader.go -f invalid-writer
env-loader: error: enum value must be one of dotenv, got 'invalid-writer', try --help
exit status 1
```

tools/env-loader/cmd/env-loader.go
fheinecke: This shouldn't occur - kingpin CLI arg validation will ensure that the key exists in the map.

```console
$ go run ./cmd/env-loader.go -f ""
env-loader: error: enum value must be one of dotenv, got '', try --help
exit status 1
$ go run ./cmd/env-loader.go -f invalid-writer
env-loader: error: enum value must be one of dotenv, got 'invalid-writer', try --help
exit status 1
```

doggydogworld: Ah true. I did a quick run before `kingpin.Parse` was added and it panic'd here but this shouldn't be an issue with parsing in place.

tools/env-loader/pkg/writers/dotenv.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Outputs the values in .env file format
+type DotenvWriter struct{}
+
+// Create a new dotenv-format writer
+func NewDotenvWriter() *DotenvWriter {
+	return &DotenvWriter{}
+}
+
+func (*DotenvWriter) validateValue(key, _ string) error {
```
doggydogworld, line 38: I think this function is small enough and only referenced once that it doesn't really need to exist. I would rather it just be inlined below.

fheinecke: It's not clear from this PR (which I split tests out of) but because this approach is easier and cleaner to test. With the validation logic (which is currently very small) in a separate function, it makes tests both easier to write and easier to understand.

I'll inline it if you still want this - let me know.

tools/env-loader/cmd/env-loader.go
```diff
@@ -0,0 +1,101 @@
...
+		return trace.Wrap(err, "failed to format output values with writer %q", c.Writer)
+	}
+
+	// Write it to stdout
+	fmt.Print(envValueOutput)
+	return nil
+}
+
+func main() {
+	c := parseCLI()
```
r0mant, line 95: We should probably do some validation on the parsed config, to make sure all required fields are set, etc. and exit early if there's an error. We usually do a `Check()` method on the config struct that validates it.

tools/env-loader/pkg/loaders/subloader.go
```diff
@@ -0,0 +1,60 @@
...
+func NewSubLoader(loaders ...Loader) *SubLoader {
+	return &SubLoader{
+		SubLoaders: loaders,
+	}
+}
+
+func (sl *SubLoader) GetEnvironmentValues(bytes []byte) (map[string]string, error) {
+	subloader := sl.getSubloader(bytes)
+	if subloader == nil {
+		return nil, trace.Errorf("found no YAML loaders for the provided content")
```
r0mant, line 36: Why YAML specifically?

tools/env-loader/pkg/loaders/yaml.go
```diff
@@ -0,0 +1,93 @@
...
+	if len(yamlBytes) == 0 {
+		return false
+	}
+
+	_, err := pys.GetEnvironmentValues(yamlBytes)
+	return err == nil
+}
+
+func (*plainYAMLSubloader) Name() string {
+	return "plain"
```
r0mant, line 54: Nit: Should this just say "YAML" or "plain YAML"?
```suggestion
	return "plain YAML"
```

tools/env-loader/pkg/envloader.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+
+		path = parent
+	}
+}
+
+// Finds environment value files for the given environment and value set, under the given directory.
+// File names will be returned in order of priority, with the lowest priority names first.
+func FindEnvironmentFilesInDirectory(environmentsDirectoryPath, environmentName, valueSet string) ([]string, error) {
+	commonFileGlob := "common.*"
```
r0mant, line 70: Nit: Move to a global constant?

tools/env-loader/pkg/envloader.go
```diff
@@ -0,0 +1,158 @@
...
+	valueSetFiles, err := filepath.Glob(filepath.Join(environmentDirectoryPath, valueSet+".*"))
+	if err != nil {
+		return nil, trace.Wrap(err, "failed to find %q value files", valueSet)
+	}
+
+	matchedFiles := append(commonFiles, valueSetFiles...)
+	if matchedFiles == nil {
+		matchedFiles = []string{}
+	}
+
```
r0mant, line 100: I don't think you need this.
```suggestion
```

tools/env-loader/pkg/envloader.go
r0mant: Yeah, we use this pattern very sparingly in main Teleport codebase, I personally don't have a strong preference either way. The current way looks fine to me too and is slightly cleaner from the caller perspective.

r0mant (approved):

tools/env-loader/cmd/env-loader.go
fheinecke: What additional checks are needed here? The minimum config is already enforced (`--environment` and `--format` flags, via a default value).

tools/env-loader/pkg/loaders/subloader.go
fheinecke: I missed this when refactoring some code. Removed the `YAML` part.

tools/env-loader/pkg/loaders/yaml.go
fheinecke: `plain YAML` is probably more informative. Updated here and with the SOPS loader

tools/env-loader/cmd/env-loader.go
fheinecke: Going to merge but will file another PR with additional checks if needed

PR #282, opened 2024-10-23
Changed environment loader to split environment names on '/'

fheinecke:
This is a minor change to the environment loader tool to allow for loading values from additional "common" files.

Previously, if an environment name contained a slash (i.e. `publish/prod`), the following file globs would be loaded:
* `<root env dir>/common.*`
* `<root env dir>/publish/prod/common.*`

This PR splits the name into components using the `/` character as a separator (on all platforms), and checks each part. Now, the following file globs would be loaded:
* `<root env dir>/common.*`
* `<root env dir>/publish/common.*`
* `<root env dir>/publish/prod/common.*`

This small but powerful change will greatly reduce the amount of duplication we have in cases [like this](https://github.com/gravitational/github-terraform/blob/main/gravitational/github-actions-values.tf#L293-L317).

### DIFF
--- a/tools/env-loader/pkg/envloader.go
+++ b/tools/env-loader/pkg/envloader.go
@@ -19,6 +19,7 @@ package envloader
 import (
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/gravitational/shared-workflows/tools/env-loader/pkg/loaders"
 	"github.com/gravitational/trace"
@@ -27,6 +28,10 @@ import (
 // This is the path, relative to the git repo root, where environment
 // directories can be found.
 const CIDirectoryRelativePath = ".environments"
+
+// Character used to separate directories in an environment name. This must be
+// consistent across multiple platforms.
+const EnvironmentNameDirectorySeparator = "/"
 
 // Glob that matches "common" files which should always be loaded. The
 // values in these files have a lower "preference" than more specific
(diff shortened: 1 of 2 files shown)


tools/env-loader/pkg/envloader.go
```diff
@@ -69,35 +74,59 @@ func findGitRepoRoot() (string, error) {
...
-		return nil, trace.Wrap(err, "failed to find common value files")
-	}
+// Given a base path (typically the environments directory) and a relative subdirectory path
+// (typically an environment name), find all common value files in every directory between
+// the two, inclusively.
+// NOTE: this assumes '/' is used a directory separator character. This is important so that
+// the same results are produced on multiple platforms.
+func findCommonFilesInPath(basePath, relativeSubdirectoryPath string) ([]string, error) {
+	relativeSubdirectoryPath = filepath.Clean(relativeSubdirectoryPath)
+	subdirectoryNames := strings.Split(relativeSubdirectoryPath, EnvironmentNameDirectorySeparator)
```
zmb3, line 84: Nit: use `path.Split` if you always want to split on `/`, or `filepath.Split` if you want to split on the OS-specific path separator.

fheinecke: > path.Split
> filepath.Split

Unfortunately these just split the last item, rather than all of them (similar to `dirname`/`basename`). Instead, I want to split every item into a list. That is, convert `"a/b/c/d"` into `[]string{"a", "b", "c", "d"}`.

Is there a function that does this? I checked both `path` and `filepath` and didn't see anything, but I might have missed something.

zmb3: Oof, I see.

Then no I think `strings.Split` is the best you can do.

If you are going to continue to split on `/` then it wouldn't hurt to do `filepath.ToSlash()` on the input on the off chance this code ever runs on a platform like Windows.

fheinecke: > on the off chance this code ever runs on a platform like Windows.

This isn't an off chance - it's a certainty. This will run on every platform that we build on. That said this is a good idea - added.

zmb3: Ok that works for me, but I think you want to invert the order. Do the `filepath.Clean` first (because it's going to operate on the OS-specific separator), and then send convert the cleaned path to a uniform separator with `filepath.ToSlash`.

The other option, which seems a bit better now that I think about it, would be to eliminate the `ToSlash` and instead do a `strings.Split(s, filepath.Separator)` and get rid of your `EnvironmentNameDirectorySeparator` constant.

tools/env-loader/pkg/envloader.go
```diff
@@ -69,35 +74,59 @@ func findGitRepoRoot() (string, error) {
...
+	relativeSubdirectoryPath = filepath.Clean(filepath.ToSlash(relativeSubdirectoryPath))
+	subdirectoryNames := strings.Split(relativeSubdirectoryPath, EnvironmentNameDirectorySeparator)
+	directoryNamesToCheck := append([]string{"."}, subdirectoryNames...)
+
+	var commonFilePaths []string
+	currentDirectoryPath := basePath
+	for _, directoryNameToCheck := range directoryNamesToCheck {
+		currentDirectoryPath := filepath.Join(currentDirectoryPath, directoryNameToCheck)
+
+		fileInfo, err := os.Lstat(currentDirectoryPath)
```
zmb3, line 92: Could anything bad happen if this is a symlink? If an attacker can put a symlink at the right place can they make us write credentials somewhere we shouldn't?

Teleport has a number of "safe open" utilities after receiving several audit findings related to symlinks.

tools/env-loader/pkg/envloader.go
fheinecke: > strings.Split(s, filepath.Separator)

I considered this, but don't think it's what we want here. When we use this tool in our build pipelines, we will do something like:

`env-loader --environment "prod/publish" --value-set gpg-signing >> "${GITHUB_ENV}"`

Ultimately the separator character is one chosen for convenience (it's the separator used in both URLs and unix system file systems), and strictly speaking isn't expected to be the OS file path separator character. IMO it would be better to keep the environment name consistent across all platforms, rather than slightly different for non-unix platforms. This should help prevent bugs when it's used in practice.

tools/env-loader/pkg/envloader.go
fheinecke: > Could anything bad happen if this is a symlink? If an attacker can put a symlink at the right place can they make us write credentials somewhere we shouldn't?

Yes I _think_ so, in some corner cases. This is why I'm using `lstat` rather than `stat`, so that if the entry is a symlink the tool doesn't follow it.

zmb3 (approved):

doggydogworld (approved):

PR #284, opened 2024-11-04
Setup env-loader CI/CD and dev tooling

fheinecke:
This PR creates dev and release tooling for the env-loader tool. This reuses tooling that was previously written for the changelog generator. The tooling includes:
* Makefile with targets for linting, testing, and building for all supported platforms
* PR testing workflow
* Release workflow
* Dev container config files for local development

While the git diff makes this unclear, the only real changes to the files previously under `tools/changelog` are very slight generalization to remove hard-coded "`changelog`" values. Same thing with the GHA workflows. The actual logic is the same, just pulled out into reusable workflows with hard-coded, tool-specific values removed.

I've also included a dep update in the PR to avoid filing an additional one, but I can split it out if requested.


.github/workflows/env-loader-ci.yaml
```diff
@@ -0,0 +1,14 @@
+---
+name: Environment value loader CI
+
+on:
+  pull_request:
```
r0mant, line 5: I understand release workflows being separate but for CI can we simplify and just have one workflow that executes all tests? Right now tools may be nicely separated but what if they say use a shared library and it gets modified, this sort of separation will just add headache later. Let's just have a single "unit tests" workflow that runs all Go tests.

r0mant (approved):

PR #285, opened 2024-11-04
Added action for env-loader

fheinecke:
This adds an action that will be used to run the environment value loader in release pipelines.

Depends on https://github.com/gravitational/shared-workflows/pull/284, #286.


r0mant (approved):

fheinecke:
Full release using this is available here: https://github.com/gravitational/teleport.e/actions/runs/12957242851/job/36311376449

PR #287, opened 2024-11-06
`amplify-preview`: custom action to trigger amplify previews

taraspos:
### Summary

This PR creates custom github action which:
- Accepts list of Amplify App IDs
- Checks all Amplify App IDs to find if branch was already connected to one of the apps
- If `create-branch` enabled
  - Connect branch to one of the AWS Amplify where limit hasn't been reached yet
  - Trigger initial deployment job
- If `wait` enabled
  - Wait for deployment to succeed/fail
- Post/update comment on the PR with preview URL

### Context
Because of limitations with AWS Amplify, we can't use Amplify's GitHub integration in `teleport` or `docs-website` repos:
1. Amplify has hard limits of 50 Amplify previews per app[^1], and we have more PRs in teleport app
    [^1]: https://docs.aws.amazon.com/general/latest/gr/amplify.html
1. There is no way to create PR preview programatically, to use existing status checks and PR comments:
    - https://github.com/aws-amplify/amplify-hosting/issues/3963
1. If PR previews are created automatically there is no way to filter for which PRs to create previews, so as results previews are created even for PRs where docs were not modified which makes previews limit  problem even more severe:
    - https://github.com/aws-amplify/amplify-hosting/issues/3960
1. Diff mechanism that exist doesn't work properly either:
    - https://github.com/aws-amplify/amplify-hosting/issues/3959
 

### Testing

Tested on:
- https://github.com/gravitational/teleport/pull/48512

This is a more advanced version of workflow running in `docs-website` repo, once this is merged, I will configure `docs-website` to run this version:
  - https://github.com/gravitational/docs-website/pull/22

Run example:
- https://github.com/gravitational/teleport/actions/runs/12353860346/job/34473886441?pr=48512

PR comment looks like following:

image

or if latest deployment is not successful, it will show latest deployment and latest successful deployment.

image


.github/actions/amplify-preview/action.yaml
```diff
@@ -0,0 +1,141 @@
+name: Amplify Preview
+description: Prepare Amplify Preview URL and post it in PR comments
+inputs:
+  app_ids:
+    description: "Comma separated list of Amplify App IDs"
+    required: true
```
fheinecke, line 6: This is actually meaningless - GHA doesn't check it for composite actions. A step in the action needs to validate it instead

.github/actions/amplify-preview/action.yaml
```diff
@@ -0,0 +1,141 @@
...
+      id: extract_branch
+
+    - name: Get or create Amplify branch
+      id: get_amplify_branch
+      env:
+        AMPLIFY_APP_IDS: ${{ inputs.app_ids }}
+        BRANCH_NAME: ${{ steps.extract_branch.outputs.branch }}
+        CREATE_BRANCHES: ${{ inputs.create_branches }}
+      shell: bash
+      run: |
```
fheinecke, line 26: Please run these through shellcheck - there are several reported issues.

.github/actions/amplify-preview/action.yaml
```diff
@@ -0,0 +1,141 @@
+name: Amplify Preview
```
fheinecke, line 1: Can you move this per [this RFD](https://github.com/gravitational/shared-workflows/blob/main/rfd/0001-tooling-requirements.md#repo-structure) and add a README?

fheinecke (commented):
IMO this should be written as a Go or typescript tool. We've been trying to move this direction to reduce our reliance on GHA.

If you decide to go this route, there are tools in [this directory](https://github.com/gravitational/shared-workflows/tree/main/tools/repo-release-tooling) for CI/CD workflows as well as local tooling.

tools/amplify-preview/go.mod
```diff
@@ -0,0 +1,38 @@
...
+	github.com/google/go-github/v63 v63.0.0 // indirect
+	github.com/google/go-querystring v1.1.0 // indirect
+	github.com/gravitational/trace v1.4.0 // indirect
+	github.com/xhit/go-str2duration/v2 v2.1.0 // indirect
+	golang.org/x/net v0.23.0 // indirect
+	golang.org/x/oauth2 v0.21.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+)
+
+replace github.com/gravitational/shared-workflows/libs => ../../libs/
```
taraspos, line 38: This is needed to use new functions added to the `github` package.
Can be removed once new version is released.

tools/amplify-preview/main.go
```diff
@@ -0,0 +1,136 @@
...
+
+import (
+	"context"
+	"errors"
+	"log/slog"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/alecthomas/kingpin/v2"
```
jimbishopp, line 27: For consistency, let's use either the flag package or cobra.

E.g.
https://github.com/gravitational/shared-workflows/blob/0f1f9dec1c978a238eae82bfa92a3d6357ad3f1e/bot/main.go#L118-L131
https://github.com/gravitational/cloud/blob/master/tcctl/release/cli/cli.go

tools/amplify-preview/main.go
```diff
@@ -0,0 +1,136 @@
...
+const (
+	jobWaitSleepTime    = 30 * time.Second
+	jobWaitTimeAttempts = 40
+)
+
+func main() {
+	kingpin.Parse()
+	ctx := context.Background()
+
+	cfg, err := config.LoadDefaultConfig(context.TODO(), config.WithRetryer(func() aws.Retryer {
```
jimbishopp, line 55: I'd use a context with cancelation. If you switch to cobra, its `Command` includes a context tied to signals.

tools/amplify-preview/main.go
```diff
@@ -0,0 +1,136 @@
...
+	amp := AmplifyPreview{
+		client: amplify.NewFromConfig(cfg),
+		appIDs: func() []string {
+			if len(*amplifyAppIDs) == 1 {
+				// kingpin env variables are separated by new lines, and there is no way to change the behavior
+				// https://github.com/alecthomas/kingpin/issues/249
+				return strings.Split((*amplifyAppIDs)[0], ",")
+			}
+			return *amplifyAppIDs
+		}(),
```
jimbishopp, line 72: Try to avoid two things: dereferencing pointers & inlined auto-executing functions. They are buggy and difficult to follow.

Cobra may handle the newlines issue better by using its `Slice` flags funcs, and pointers are not needed when using its `Var` flag functions. That may remove the need for the inlined func call, otherwise I'd just define a new variable above this AmplifyPreview init that creates the app ids slice and when assign it here.

tools/amplify-preview/main.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	// Check if Amplify branch is already connected to one of the Amplify Apps
+	branch, err := amp.FindExistingBranch(ctx, *gitBranchName)
+	if err != nil {
+		if !*createBranches && !errors.Is(err, errNoJobForBranch) {
+			logger.Error("failed to lookup branch", logKeyBranchName, *gitBranchName, "error", err)
+			os.Exit(1)
```
jimbishopp, line 80: As a general practice, exiting in the middle of execution is dangerous. Not so much at here, but some other engineer may decide to open a connection above this that wouldn't get closed. Better to define a `run() error` function so the errors can be returned from here and call it from `main`. 

Cobra would provide similar behavior: https://github.com/gravitational/cloud/blob/master/tcctl/cmd/tcctl/main.go

tools/amplify-preview/main.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		// if job not found and branch was just created - start new job
+		latestJob, err = amp.StartJob(ctx, branch)
+		if err != nil {
+			logger.Error("failed to start amplify job", logKeyBranchName, *gitBranchName, "error", err)
+			os.Exit(1)
+		}
+	}
```
jimbishopp, line 105: Consider splitting all these steps into multiple functions to help simplify the workflow and simplify readability. A lot of nesting makes the code difficult to follow. And the multiple statements mean it's more difficult to verify scoping of variables is accurate.

tools/amplify-preview/main.go
```diff
@@ -0,0 +1,136 @@
...
+
+	if err := postPreviewURL(ctx, amplifyJobsToMarkdown(branch, latestJob, activeJob)); err != nil {
+		logger.Error("failed to post preview URL", "error", err)
+		os.Exit(1)
+	}
+	logger.Info("Successfully posted PR comment")
+
+	if *wait {
+		for i := 0; !isAmplifyJobCompleted(latestJob) && i < jobWaitTimeAttempts; i++ {
```
jimbishopp, line 114: E.g. from above comment. This would be easier to understand and follow if it said `if wait { DoSomethingElseWithALiterateName }`. Self-documenting code (with short names) is the best kind of code.

jimbishopp (commented):
👍 First pass.

tools/amplify-preview/action.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - name: Amplify Preview
+      env:
+        AMPLIFY_APP_IDS: ${{ inputs.app_ids }}
+        GIT_BRANCH_NAME: ${{ steps.extract_branch.outputs.branch }}
+        CREATE_BRANCHES: ${{ inputs.create_branches }}
+        GITHUB_TOKEN: ${{ inputs.github_token }}
+        WAIT: ${{ inputs.wait }} 
+      shell: bash
+      run: |
+        pushd ./tools/amplify-preview/; go run ./; popd
```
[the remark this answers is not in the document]
taraspos: This can be removed in subsequent release PR once `libs` package is released and I can remove the `replace` rule from go.mod:


https://github.com/gravitational/shared-workflows/blob/b7d78994a6856a6c8f1959fe974def5810b2aa9f/tools/amplify-preview/go.mod#L37

tools/amplify-preview/main.go
taraspos: All tools in https://github.com/gravitational/shared-workflows/tree/master/tools are using `github.com/alecthomas/kingpin` and I followed their example for consistency 😬

libs/github/comment.go
```diff
@@ -0,0 +1,69 @@
...
+	Owner  string
+	Repo   string
+	Number int
+}
+
+// every trait is treated as AND
+type CommentTraits struct {
+	BodyContains *string
+	UserLogin    *string
+}
```
jimbishopp, line 25: Forgot to comment on this. Let's avoid using pointers like this. I know the AWS Go client does this but it is really bad practice. If we need to track three states (has value, zero value, nil value) then it is better to create a specific type. The `Null*` values in the [sql package](https://pkg.go.dev/database/sql#NullBool) are an example of this.

Defining useful zero values is a Go idiom that prevents all kinds of bugs and simplifies code. See [Go Proverbs](https://go-proverbs.github.io/), specifically the excellent talk by Rob Pike: [Make the zero value useful.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=6m25s).

tools/amplify-preview/main.go
jimbishopp: Not sure why kingpin was chosen for those tools. The review bot uses the simple flag package. We previously used kingpin in cloud but changed to cobra because it has a better design.

taraspos: I think it's because teleport uses kingpin everywhere:
- https://github.com/search?q=repo%3Agravitational%2Fteleport%20kingpin&type=code

libs/github/comment.go
taraspos: After some consideration I decided that I can just use plain strings in this case.
Because matching body content or user login against empty strings makes no sense, so we can treat empty string as zero value.

tools/amplify-preview/github.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+const (
+	prRefNameSuffix = "/merge"
+)
+
+func postPreviewURL(ctx context.Context, commentBody string) error {
+	refName := os.Getenv("GITHUB_REF_NAME")
+	githubRepository := os.Getenv("GITHUB_REPOSITORY")
+	if !strings.HasSuffix(refName, prRefNameSuffix) {
```
jimbishopp, line 38:
```suggestion
func postPreviewURL(ctx context.Context, commentBody string) error {
	const prRefNameSuffix = "/merge"
	refName := os.Getenv("GITHUB_REF_NAME")
	githubRepository := os.Getenv("GITHUB_REPOSITORY")
	if !strings.HasSuffix(refName, prRefNameSuffix) {
```

tools/amplify-preview/github.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	gh, err := github.NewClientFromGHAuth(ctx)
+	if err != nil {
+		return err
+	}
+
+	prID, err := strconv.Atoi(strings.TrimSuffix(refName, "/merge"))
+	if err != nil {
+		log.Fatalf("Failed to extract PR ID from GITHUB_REF_NAME=%s: %s", refName, err)
```
jimbishopp, line 49: Why panic?

tools/amplify-preview/github.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	targetComment := github.CommentTraits{
+		BodyContains: amplifyMarkdownHeader,
+	}
+
+	currentPR := github.IssueIdentifier{
+		Number: prID,
+		Owner:  strings.Split(githubRepository, "/")[0],
+		Repo:   strings.Split(githubRepository, "/")[1],
```
jimbishopp, line 59: Split once. Also wouldn't risk a panic even though it is unlikely.

tools/amplify-preview/main.go
```diff
@@ -0,0 +1,150 @@
...
+	wait = kingpin.Flag("wait",
+		"Wait for pending/running job to complete").Envar("WAIT").Default("false").Bool()
+)
+
+const (
+	jobWaitSleepTime    = 30 * time.Second
+	jobWaitTimeAttempts = 40
+)
+
+func main() {
```
jimbishopp, line 52: Recommend the `run() error` idiom where main only calls run and logs an error + sets exit code if a non-nil error is returned from run. One log error statement and `os.Exit(1)` call.

tools/amplify-preview/main.go
```diff
@@ -0,0 +1,150 @@
...
+	kingpin.Parse()
+	ctx := context.Background()
```
jimbishopp, line 54: Context here should cancel on os signal. You'll need to track signals since kingpin doesn't support it.

tools/amplify-preview/amplify.go
```diff
@@ -0,0 +1,338 @@
...
+			logger.Debug("Branch not found", logKeyAppID, resp.appID, logKeyBranchName, amp.branchName)
+			continue
+		} else if resp.err != nil {
+			failedResp.perAppErr[resp.appID] = resp.err
+			continue
+		}
+
+		if resp.data != nil {
+			return resp.data.Branch, nil
+		}
```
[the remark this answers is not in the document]
taraspos: It should be always just one branch, but we don't know which app it's connected to.
We totally can iterate without goroutines, just thought this will save save second or two, so we don't have to wait on sequential responses.

tools/amplify-preview/amplify.go
```diff
@@ -0,0 +1,338 @@
...
+			AppId:           aws.String(appID),
+			BranchName:      aws.String(amp.branchName),
+			Description:     aws.String("Branch generated for PR TODO"),
+			Stage:           types.StagePullRequest,
+			EnableAutoBuild: aws.Bool(true),
+		})
+
+		var errLimitExceeded *types.LimitExceededException
+		if errors.As(err, &errLimitExceeded) {
+			logger.Debug("Reached branches limit", logKeyAppID, appID)
```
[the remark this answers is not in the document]
taraspos: Reaching limit is expected behavior in this case.
Because amplify has limit of 50 branches per app, if first app is full we want to move to the next one where there still room left.

tools/amplify-preview/amplify.go
```diff
@@ -0,0 +1,338 @@
...
+		}()
+	}
+
+	wg.Wait()
+	close(resultCh)
+
+	failedResp := aggregatedError{
+		perAppErr: map[string]error{},
+		message:   "failed to fetch branch",
+	}
```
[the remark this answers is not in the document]
taraspos: The original purpose was not really to deduplicate errors but just to associate error with app where it happened, and then combine in one aggregated error.

tools/amplify-preview/amplify.go
```diff
@@ -0,0 +1,338 @@
...
+
+	return jobSummaries, nil
+}
+
+func (amp *AmplifyPreview) GetLatestAndActiveJobs(ctx context.Context, branch *types.Branch) (latestJob, activeJob *types.JobSummary, err error) {
+	var jobIDs []string
+	if branch.ActiveJobId != nil {
+		jobIDs = append(jobIDs, *branch.ActiveJobId)
+	}
+	jobSummaries, err := amp.findJobsByID(ctx, branch, true, jobIDs...)
```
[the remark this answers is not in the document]
taraspos: There could be case when `ActiveJobId` is nil, so it will be empty slice. This way it allows passing one more no IDs.

tools/amplify-preview/main.go
taraspos: I would prefer to keep it as is to avoid spending more time on this, as long term we would prefer using native integration, once Amplify get to resolve current issues.

jimbishopp (approved):

PR #289, opened 2024-11-07
Add a workflow to enforce new docs redirects

ptgott:
Add a `docpaths` workflow to the bot to ensure that any renamed or deleted docs pages accompany a new redirect in the docs config file (`config.json`). This way, we can prevent docs changes from introducing 404s, whether because of delays in search engine indexing or broken links from Teleport-owned sites.

The workflow takes a path to a `config.json` file, loads the redirect configuration, and checks whether all renamed and deleted pages correspond to a redirect. This change adds a `docpaths` value of the `workflow` flag and an optional `teleport-path` flag for the path to a `gravitational/teleport` clone so the workflow can locate a docs configuration file.


ptgott:
Here is an example of a failing workflow run in a v16 test PR (I changed the error message a bit since this workflow run): https://github.com/gravitational/teleport/actions/runs/11726069638/job/32663859094?pr=48546#step:7:18

bot/internal/bot/docpaths.go
```diff
@@ -0,0 +1,122 @@
...
+// directory of a Teleport repo. We only use Redirects, so other fields have the
+// "any" type.
+type DocsConfig struct {
+	Navigation any
+	Variables  any
+	Redirects  RedirectConfig
+}
+
+type RedirectConfig []Redirect
+
```
zmb3, line 27: Is the extra type necessary?

```suggestion
	Redirects  []Redirect
}


```

bot/internal/bot/docpaths.go
```diff
@@ -0,0 +1,122 @@
...
+type Redirect struct {
```
zmb3, line 28: This will be referred to as `bot.Redirect` - which doesn't make it super obvious that it's related to docs.

I would either have a `docs` package so this can be named `docs.Redirect` or rename this to something like `DocsRedirect`.

bot/internal/bot/docpaths.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// CheckDocsPathsForMissingRedirects takes the relative path to a
+// gravitational/teleport clone. It assumes that there is a file called
+// "docs/config.json" at the root of the directory that lists redirects in the
+// "redirects" field.
+//
+// CheckDocsPathsForMissingRedirects checks whether a PR has renamed or deleted
+// any docs files and, if so, returns an error if any of these docs files does
+// not correspond to a new redirect in the configuration file.
```
zmb3, line 41: Looks like two separate descriptions here - did you mean to keep both paragraphs or was one meant to replace the other?

bot/internal/bot/docpaths.go
```diff
@@ -0,0 +1,122 @@
...
+func (b *Bot) CheckDocsPathsForMissingRedirects(ctx context.Context, teleportClonePath string) error {
+	if teleportClonePath == "" {
+		return trace.Wrap(errors.New("unable to load Teleport documentation config with an empty path"))
```
zmb3, line 44:
```suggestion
		return trace.BadParameter("unable to load Teleport documentation config with an empty path")
```

bot/internal/bot/docpaths.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	docsConfigPath := path.Join(teleportClonePath, "docs", "config.json")
```
zmb3, line 47:
```suggestion
	docsConfigPath := filepath.Join(teleportClonePath, "docs", "config.json")
```

bot/internal/bot/docpaths.go
```diff
@@ -0,0 +1,122 @@
...
+	f, err := os.Open(docsConfigPath)
+	if err != nil {
+		return trace.Wrap(fmt.Errorf("unable to open docs config file at %v: %v", docsConfigPath, err))
+	}
+	defer f.Close()
+
+	var c DocsConfig
+	if err := json.NewDecoder(f).Decode(&c); err != nil {
+		return trace.Wrap(fmt.Errorf("unable to load redirect configuration from %v: %v", docsConfigPath, err))
```
zmb3, line 56: Here and below, try to avoid creating two errors at once. The inner `fmt.Errorf` is rarely necessary.

```suggestion
		return trace.Wrap(err, "unable to load redirect configuration from %v: %v", docsConfigPath)
```

bot/internal/bot/docpaths.go
```diff
@@ -0,0 +1,122 @@
...
+
+	m := missingRedirectSources(c.Redirects, files)
+	if len(m) > 0 {
+		return trace.Wrap(fmt.Errorf("docs config at %v is missing redirects for the following renamed or deleted pages: %v", docsConfigPath, strings.Join(m, ",")))
+	}
+
+	return nil
+}
+
+const docsPrefix = "docs/pages/"
```
zmb3, line 72: `filepath.Join("docs", "pages")` would be best practice, but I'm okay assuming this action will always run on a system with a `/` path separator.

bot/internal/bot/docpaths.go
```diff
@@ -0,0 +1,122 @@
...
+	d, f := filepath.Split(p)
+
+	// The page is a category index page, since the name of the file is the
+	// same as the name of the containing directory. This means that the
+	// route Docusaurus generates for the page is the path to the containing
+	// directory, omitting the final path segment.
+	// See:
+	// https://docusaurus.io/docs/sidebar/autogenerated#category-index-convention
+	if strings.HasSuffix(d, strings.Replace(f, ".mdx", "/", 1)) {
+		p = d
```
zmb3, line 86: This function is a bit hard to grok - maybe it's worth making the variable names longer than single characters.

zmb3 (approved):

PR #291, opened 2024-11-11
Add a workflow to check docs URLs

ptgott:
Like the workflow to check for necessary redirects to match renamed or deleted docs pages, this workflow takes the path to a `gravitational/teleport` clone. It finds mentions of Teleport docs site paths in each file changed by a PR and ensures that each path has a corresponding docs page or redirect source.

### DIFF
--- a/bot/go.mod
+++ b/bot/go.mod
@@ -5,6 +5,7 @@ go 1.23
 require (
 	github.com/google/go-github/v37 v37.0.0
 	github.com/gravitational/trace v1.5.0
+	github.com/spf13/afero v1.12.0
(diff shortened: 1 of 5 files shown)


ptgott:
This PR is based on #289.

An example of a failed run is here: https://github.com/gravitational/teleport/actions/runs/11784998385/job/32825325002?pr=48761#step:7:8

ptgott:
Since we're going to be ensuring that renamed/deleted docs pages have redirects anyway with the `docpaths` workflow, and have been checking for inaccurate docs URLs outside CI, I think we can close this for now.

PR #292, opened 2024-11-12
Fix tooling release issues

fheinecke:
The env-loader release is failing due to this. Trim `/` from the version, and check only the `VERSION` to determine if a release is a prerelease or not.

### DIFF
--- a/.github/workflows/reusable-cd.yaml
+++ b/.github/workflows/reusable-cd.yaml
@@ -38,7 +38,7 @@ jobs:
             # should move into this repo and replace this logic
             echo "should-release=true" >> "${GITHUB_OUTPUT}"
             
-            # Any '-' character means in the version means that it is a prerelease
+            # Any '-' character in the version means that it is a prerelease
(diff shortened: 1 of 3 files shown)


.github/workflows/reusable-cd.yaml
```diff
@@ -31,14 +31,15 @@ jobs:
...
+            VERSION="${GITHUB_REF#refs/tags/tools/${TOOL_NAME}/}"
+            echo "version=${VERSION}" >> "${GITHUB_OUTPUT}"
 
             # Eventually the parse-version action from the teleport.e repo
             # should move into this repo and replace this logic
             echo "should-release=true" >> "${GITHUB_OUTPUT}"
             
-            # Any '-' character means in a tag ref means that it is a prerelease
-            if [[ "${GITHUB_REF}" == *-* ]]; then
+            # Any '-' character means in the version means that it is a prerelease
```
r0mant, line 41:
```suggestion
            # Any '-' character in the version means that it is a prerelease
```

r0mant (approved):

PR #294, opened 2024-11-18
Change the way docs approvals work

zmb3:
Docs PRs are consistently delayed due to the the fact that we currently require two approvals to come from a small team of people who are all very busy.

To mitigate this, we will treat PRs that touch docs similarly to PRs that touch code. The new behavior is:

1. PRs continue to require two approvals, whether they touch code, docs, or both.
2. At least one of these two approvals must come from a code "owner" (otherwise known as a "group 1 reviewer"). This becomes the case whether the PR touches docs or not.
3. If the PR does not touch docs, then approvals from the docs reviewers don't count towards the 2 approval limit.
4. If the PR does touch docs, then the docs reviewers are added to group 2 reviewers. Their reviews count towards the two review limit, but their reviews are not required in order to merge the PR.

### DIFF
--- a/bot/internal/review/review.go
+++ b/bot/internal/review/review.go
@@ -435,31 +435,15 @@ func (r *Assignments) CheckInternal(e *env.Environment, reviews []github.Review,
 		return nil
 	}
 
-	switch {
-	case changes.Docs && changes.Code:
-		log.Printf("Check: Found docs and code changes.")
-		if err := r.checkInternalReviews(e, changes, reviews, files); err != nil {
-			return trace.Wrap(err)
-		}
-	case !changes.Docs && changes.Code:
-		log.Printf("Check: Found code changes.")
-		if err := r.checkInternalReviews(e, changes, reviews, files); err != nil {
-			return trace.Wrap(err)
-		}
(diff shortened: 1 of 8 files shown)


bot/internal/review/review.go
```diff
@@ -438,20 +438,17 @@ func (r *Assignments) CheckInternal(e *env.Environment, reviews []github.Review,
 	switch {
 	case changes.Docs && changes.Code:
 		log.Printf("Check: Found docs and code changes.")
-		if err := r.checkInternalDocsReviews(e, reviews, files); err != nil {
-			return trace.Wrap(err)
-		}
-		if err := r.checkInternalCodeReviews(e, changes, reviews); err != nil {
+		if err := r.checkInternalReviews(e, changes, reviews, files); err != nil {
```
jimbishopp, line 441: nbd - code is easy to follow - but this same call is repeated in 3 of the 4 cases. I did have to look closely to see that the first three cases made the same call. One option would be to just log in the 3 cases, return from the 4th admin check, then make this call below the switch.

jimbishopp (approved):
Looks right.

PR #295, opened 2024-11-19
env-loader: bug fixes

fheinecke:
Fixed the following bugs with the env loader tool:
* `environments-directory` arg shared an environment variable with the `environments` arg
* No output would be produced if no values were specifically requested, instead of outputting all loaded values
* Git repo root not properly detected when the repo is a submodule
* Environment names with a `/` character in them would search individual components instead of the full path (i.e. `.environments/stage` and `.environments/build` instead of `.environments/stage/build`)

### DIFF
--- a/tools/env-loader/cmd/env-loader.go
+++ b/tools/env-loader/cmd/env-loader.go
@@ -43,7 +43,7 @@ func parseCLI() *config {
 
 	kingpin.Flag("environments-directory", "Path to the directory containing all environments, defaulting to the repo root").
 		Short('d').
-		Envar(EnvVarPrefix + "ENVIRONMENT").
+		Envar(EnvVarPrefix + "ENVIRONMENTS_DIRECTORY").
 		StringVar(&c.EnvironmentsDirectory)
 
 	kingpin.Flag("environment", "Name of the environment containing the values to load").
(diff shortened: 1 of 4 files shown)


zmb3:
Is it possible to write some tests to lock in this behavior?

r0mant (approved):
lgtm assuming we add tests

tools/env-loader/cmd/env-loader.go
```diff
@@ -30,6 +31,9 @@ import (
 
 const EnvVarPrefix = "ENV_LOADER_"
 
+// This is a package-level var to assist with capturing stdout in tests
+var outputPrinter = fmt.Print
```
zmb3, line 35: It's more common to have a package level that is an `io.Writer` for where the output goes (`os.Stdout`, a `bytes.Buffer`, etc), as opposed to making the var the function that is used.

tools/env-loader/cmd/env-loader_test.go
```diff
@@ -0,0 +1,135 @@
...
+				"topLevelCommon1": "top level",
+				"topLevelCommon2": "env level",
+			},
+		},
+	}
+
+	for _, test := range tests {
+		actualValues, err := getRequestedEnvValues(test.c)
+		require.NoError(t, err)
+		require.EqualValues(t, actualValues, test.expectedValues)
```
zmb3, line 90: I think you've got the order backwards here. I'm pretty sure it should be expected values, then actual values.

In Teleport we have a linter enabled that catches a lot of these.

tools/env-loader/cmd/env-loader.go
fheinecke: Thanks - that's a much cleaner approach

tools/env-loader/cmd/env-loader_test.go
fheinecke: > In Teleport we have a linter enabled that catches a lot of these.

Do you happen to know what it is/how it's configured? I setup [this](https://github.com/gravitational/shared-workflows/blob/main/tools/repo-release-tooling/tooling.mk#L25) to standardize how we lint across the projects in this repo. It'd be great to get this enabled for everything.

zmb3: We run `testifylint` via `golangci-lint`: https://github.com/gravitational/teleport/blob/master/.golangci.yml#L252-L264

PR #299, opened 2024-12-17
Add support for masking secret values in a GHA context

fheinecke:
This adds support for masking secret values when the tool is ran in a GHA pipeline.

This can be broken down into two parts:
* Rather than treating values as strings, they're now treated as a new `values.Value` type. This type contains not only the raw value, but also an indication as to whether or not it should be treated as a secret. All values from SOPS YAML files are now marked as `ShouldMask`.
* A new GHA-specific writer was added that outputs secret values with the `::add-mask::` prefix. [Per the docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#masking-a-value-in-a-log), this prefix causes the subsequent string (until line end) to be masked from GHA logs.

> [!CAUTION]
> It is security-critical that all corner cases are covered. Please check the tests thoroughly, specifically for the SOPS loader and the GHA mask writer.

This is a large PR by line count, but the vast majority of changes are either:
* tests (this is critical to prevent leaking values so the tests are pretty extensive)
* changing the type used for actual values from a string to a new struct (this touches a lot of places but is logically simple)


r0mant (commented):
Code-wise looks fine but my concern is this is yet another GHA-specific thing we now make use of. Is there a simpler way to accomplish this without having to rely on GHA-specific logic to mask things? For example, can we just not log sensitive data at all?

tools/env-loader/pkg/values/valuetype.go
```diff
@@ -0,0 +1,10 @@
+package values
+
+type Value struct {
+	UnderlyingValue string
+	ShouldMask      bool
+}
+
+func (v *Value) String() string {
+	return v.UnderlyingValue
+}
```
r0mant, line 10: Add godocs.

tools/env-loader/pkg/writers/gha-mask.go
```diff
@@ -0,0 +1,84 @@
...
+
+// Create a new GHA mask writer
+func NewGHAMaskWriter() *GHAMaskWriter {
+	return &GHAMaskWriter{}
+}
+
+func (ew *GHAMaskWriter) FormatEnvironmentValues(values map[string]values.Value) (string, error) {
+	renderedValues := make([]string, 0, len(values))
+	for key, value := range values {
+		if key == "" {
```
r0mant, line 43: This seems strange that the key emptiness check is inside this specific writer. Is there a way to validate the key validity when loading the value rather than in each specific writer?

fheinecke: While this conceptually makes, I think it would add some unnecessary complexity. Here's what I've done instead:
* Moved all the logic from the `if` statement to the `String()` function
* Added tests for _all_ writers to verify that they contain logic that covers this case

All writers still need to handle the case of `key == ""`, but they can do it in their own way, without additional complexity.

r0mant (approved):

PR #303, opened 2024-12-19
env-loader: Added gha-env writer

fheinecke:
This adds a new writer specifically for writing to `GITHUB_ENV` files. [This format](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-environment-variable) is _close_ to dotenv files, but doesn't actually match it for multiline strings.

Here's an example output:
```
key1<<EOF
multiline
value1

EOF
key2=value2
key3<<EOF
multiline
value3
EOF

```

Here's what it looks like for dotenv:
```
key1=multiline
value1

key2=value2
key3=multiline
value3
```

As shown above, dotenv doesn't handle new lines at all, and errors in this case as shown [here](https://github.com/gravitational/teleport.e/actions/runs/12421540569/job/34681591637#step:4:134).

While this is more GHA-specific logic, it is a necessary evil. We'll need something like this for _any_ pipeline service.

### DIFF
--- a/tools/env-loader/pkg/writers/gha-env.go
+++ b/tools/env-loader/pkg/writers/gha-env.go
@@ -1,0 +1,103 @@
+/*
+ *  Copyright 2024 Gravitational, Inc
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package writers
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/google/uuid"
+	"github.com/gravitational/shared-workflows/tools/env-loader/pkg/values"
+	"github.com/gravitational/trace"
+)
+
+const delimiterPrefix = "EOF"
+
+// Outputs values in a format that can be parsed by GHA's `GITHUB_ENV` file.
+// This is _almost_ the same as dotenv files, but also handles multiline
+// environment values. For details, see
+// https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-environment-variable
+type GHAEnvWriter struct{}
+
+// Create a new GHA env writer
+func NewGHAEnvWriter() *GHAEnvWriter {
+	return &GHAEnvWriter{}
+}
+
+// Generates a delimiter that is guaranteed to not contain the provided string.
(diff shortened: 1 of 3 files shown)


tools/env-loader/pkg/writers/gha-env.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Create a new GHA env writer
+func NewGHAEnvWriter() *GHAEnvWriter {
+	return &GHAEnvWriter{}
+}
+
+// Generates a delimiter that is guaranteed to not contain the provided string.
+// This is required for writing multiline values to `GITHUB_ENV`, per
+// https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#multiline-strings
+func generateMultilineDelimiter(value string) string {
```
r0mant, line 44: Since you're using UUID in your delimiter string, the chance of collision is negligible and this function is needlessly complicated and can be reduced to literally just

```go
func getDelimiter() string {
    return delimiterPrefix + "_" + uuid.NewString()
}
```

Maybe even no need for the function at all and just do it inline.

tools/env-loader/pkg/writers/gha-env.go
```diff
@@ -0,0 +1,103 @@
...
+	renderedValues := make([]string, 0, len(values))
+	for key, value := range values {
+		if key == "" {
+			return "", trace.Errorf("found empty key for log value %q", value.String())
+		}
+
+		// Don't format strings without new lines as multiline. This would be valid, but a little
+		// less readable.
+		var renderedValue string
+		if strings.Contains(value.UnderlyingValue, "\n") {
```
r0mant, line 81: Worth simplifying the logic and just using the multi-line format for everything?

r0mant (commented):
I think the code can be greatly simplified.

fheinecke:
Sure, I can simplify this as above. It'll just make the output much more difficult to visually parse when it's actually used. For example, it will turn this output:

```
key1=value1
key2=value2
key3=value3
..
# Repeat a ~10 more times for some pipelines
```

into

```
key1<<EOF_4D48B7F5-0DE7-4B8D-9FAE-6B54D4F46214
value1
EOF_4D48B7F5-0DE7-4B8D-9FAE-6B54D4F46214
key2<<EOF_EDA7F489-F228-4E33-83D4-421CA2DF366E
value2
EOF_EDA7F489-F228-4E33-83D4-421CA2DF366E
key3<<EOF_E0577A69-8A63-4EA5-A4B1-30CD5A5D7923
value3
EOF_E0577A69-8A63-4EA5-A4B1-30CD5A5D7923
# Repeat a ~10 more times for some pipelines

fheinecke:
@r0mant let me know what you think and I'll get the tests fixed. Using UUIDs everywhere every time will add an lot of boilerplate + regex to the tests, so I wanted to get feedback before I spend time on it.

r0mant (approved):
@fheinecke Yeah it does look kinda ugly (which in hindsight it was kinda obvious it would). Up to you if you want to revert that particular change, lgtm either way.

fheinecke:
I'm going to revert. If nothing else, this means that I can merge and move forward with the GHA env security project without spending more time on tests.

PR #309, opened 2025-01-17
Updating release-notes tool to account for new changelog format

doggydogworld:
The format of our CHANGELOG has changed a bit for release branches. Previously, the latest version of the release branch was expected to be the first second-level header.

e.g. for branch/v16 with the latest release being v16.0.1 the first second-level header encountered would be `## 16.0.1`

The format of the CHANGELOG for our release branches now includes all released versions. Now the first second-level header will be the latest released version of Teleport and the patches for other releases will be located further down in the CHANGELOG.

To account or this the release-notes tool will now search for the correct version in the CHANGELOG to populate the release notes.


tools/release-notes/testdata/test-changelog.md
```diff
@@ -21,3 +27,17 @@ Teleport 16 brings the following new features and improvements:
 - Access requests from the resources view
 - `tctl` for Windows
 - Teleport plugins improvements
+
+## 15.4.17 (08/28/24)
```
camscale, line 31: Perhaps add a test for this version so that there is a test where the changelog is terminated by EOF and not the next heading? Or just remove this section so that the test for v16 does that?

camscale (approved):
LGTM, but...

I'm not sure this is going to be needed because I think the changelogs are now unmaintainable so I would want to revert the change that adds the later changelogs to an earlier release branch.

That said, there is no harm to this change - maybe it may be useful to generate an old github release for some reason?

PR #321, opened 2025-02-05
Fix false negatives in the docpaths workflow

ptgott:
In `CheckDocsPathsForMissingRedirects`, a `trace.Wrap` call mistakenly includes an unrelated error as an argument, and the error is either nil or handled earlier in the function. This means that docs pull requests with insufficient redirects still cause the workflow to pass.

This change fixes the issue and adds testing for error handling in `CheckDocsPathsForMissingRedirects`. As part of this, it improves some other error messages returned by the function.

### DIFF
--- a/bot/internal/bot/docpaths.go
+++ b/bot/internal/bot/docpaths.go
@@ -3,13 +3,12 @@ package bot
 import (
 	"context"
 	"encoding/json"
-	"errors"
-	"os"
 	"path/filepath"
 	"strings"
 
 	"github.com/gravitational/shared-workflows/bot/internal/github"
 	"github.com/gravitational/trace"
+	"github.com/spf13/afero"
 )
 
 // DocsConfig represents the structure of the config.json file found in the docs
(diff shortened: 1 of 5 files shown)


bot/internal/bot/docpaths.go
```diff
@@ -3,13 +3,12 @@ package bot
...
 	"context"
 	"encoding/json"
-	"errors"
-	"os"
 	"path/filepath"
 	"strings"
 
 	"github.com/gravitational/shared-workflows/bot/internal/github"
 	"github.com/gravitational/trace"
+	"github.com/spf13/afero"
```
zmb3, line 11: Can we use `io/fs` instead of pulling in a dependency?

zmb3 (approved):

bot/internal/bot/docpaths.go
ptgott: Done in [9c4ac80](https://github.com/gravitational/shared-workflows/pull/321/commits/9c4ac80c776a7d7aac9e20729d9d35a798692d8f)

camscale (approved):

PR #324, opened 2025-02-07
Tooling for Mac Distribution

doggydogworld:


tools/mac-distribution/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+# Package
+mac-distribution package-pkg --install-location /usr/local/bin "${STAGING_PKG}" "my-app.pkg"
+```
+
+### Notarization
+
+By default, notarization is disabled and will output dryrun logs. To enable it you must either set the following options:
+```shell
+mac-distribution --apple-username="" --apple-password="" --signing-identity="" --bundle-id="" ...
```
zmb3, line 27: Should we support flags at all? Doesn't this just make it more likely that secrets get recorded in shell history files?

tools/mac-distribution/README.md
```diff
@@ -0,0 +1,42 @@
...
+APPLE_PASSWORD=""
+SIGNING_IDENTITY=""
+BUNDLE_ID=""
+```
+
+If all of these are set then notarization will be enabled and the tool will notarize after packaging.
+This is to make it convenient to test locally without having to set up creds to build packages.
+
+However this isn't desirable in CI environments where notarization must happen. Enabling dryrun will "silently" cause a failure.
+For convenience the `--force-notarization` flag is provided to fail in the scenario where creds are missing.
```
zmb3, line 42: Most CI environments set a `CI` environment variable - what do you think about automatically forcing notarization when `$CI` is set?

tools/mac-distribution/main.go
```diff
@@ -0,0 +1,141 @@
...
+		return trace.Wrap(err)
+	}
+	return trace.Wrap(tool.NotarizeBinaries(c.Files))
+}
+
+func (g *GlobalFlags) InitNotaryTool() (*notarize.Tool, error) {
+	// Dry run if no credentials are provided
+	dryRun := g.AppleUsername == "" || g.ApplePassword == "" || g.SigningID == ""
+	if dryRun && g.ForceNotarization {
+		return nil, fmt.Errorf("notarization credentials not provided and force-notarization is enabled")
```
zmb3, line 120:
```suggestion
		return nil, errors.New("notarization credentials not provided and force-notarization is enabled")
```

tools/mac-distribution/notarize/notarize.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/gravitational/shared-workflows/tools/mac-distribution/internal/zipper"
+
+	"github.com/gravitational/trace"
+)
+
+// Tool is a wrapper around the MacOS codesigning/notarizing utilities.
+type Tool struct {
+	Creds Creds
+	// Retry sets the max number of attempts for a succesful notarization
+	retry int
```
zmb3, line 19:
```suggestion
	maxRetries int
```

A well-named variable often eliminates the need for extra comments.

tools/mac-distribution/notarize/notarize.go
```diff
@@ -0,0 +1,221 @@
...
+
+	log *slog.Logger
+
+	dryRun    bool
+	cmdRunner exec.CommandRunner
+	zip       zipper.DirZipper
+}
+
+// Creds is
```
zmb3, line 28: Creds is ......

tools/mac-distribution/notarize/submit.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"encoding/json"
+
+	"github.com/gravitational/trace"
+)
+
+// SubmissionResponseData contains information about the status of a submission.
+//
+// Reference: https://developer.apple.com/documentation/notaryapi/submissionresponse/data-data.dictionary
+type SubmissionResponseData struct {
```
zmb3, line 12: Does this need to be exported?

tools/mac-distribution/packaging/appbundle.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	return nil
+}
+
+func (a *AppBundleInfo) validate() error {
+	// Validate skeleton
+	info, err := os.Stat(a.Skeleton)
+	if err != nil {
+		return trace.Wrap(err, "skeleton directory does not exist")
```
zmb3, line 88: Two things:
- should we check if `os.IsNotExist(err)`? Stat can error in other ways than just a missing file.
- Should the error message include the name of the directory we tried to stat?

tools/mac-distribution/packaging/appbundle.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	if !info.IsDir() {
+		return fmt.Errorf("skeleton must be a directory")
```
zmb3, line 92:
```suggestion
		return errors.New("skeleton must be a directory")
```

Use `errors.New` if you just have a string literal. The `fmt` version is only necessary when you have format verbs that you want to do substitution with.

tools/mac-distribution/packaging/packageinstaller.go
```diff
@@ -0,0 +1,169 @@
...
+	if p.notaryTool != nil {
+		if err := p.notaryTool.NotarizePackageInstaller(p.Info.OutputPath, p.Info.OutputPath); err != nil {
+			return trace.Wrap(err)
+		}
+	}
+
+	return nil
+}
+
+// nonRelocatablePlist analyzes the root path to create a compponent plist file.
```
zmb3, line 121:
```suggestion
// nonRelocatablePlist analyzes the root path to create a component plist file.
```

tools/mac-distribution/packaging/packageinstaller_test.go
```diff
@@ -0,0 +1,64 @@
...
+// Test only runs on macOS and requires root privileges.
+func TestPackageInstallerPackager(t *testing.T) {
+	// TestPackageInstaller tests the package installer.
+	// It installs a package and checks that the expected files are installed.
+	// It then uninstalls the package and checks that the expected files are removed.
+	bundleID := "com.gravitational.testingpackage"
+
+	testdataPath, err := filepath.Abs("testdata")
+	require.NoError(t, err)
+	tempOutdir, err := os.MkdirTemp(testdataPath, "pkgstaging")
```
zmb3, line 21: Should we use `t.Tempdir()` here?

tools/mac-distribution/internal/exec/exec.go
```diff
@@ -0,0 +1,61 @@
...
+func (d *DefaultCommandRunner) RunCommand(path string, args ...string) (string, error) {
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+
+	cmd := exec.Command(path, args...)
+	cmd.Stderr = &stderr
+	cmd.Stdout = &stdout
+
+	err := cmd.Run()
+	out := strings.TrimSpace(stdout.String())
```
zmb3, line 35: You could save some allocations by keeping a `[]byte` until the last possible moment. The `bytes` package and the `strings` package share most of these utilities in common.

tools/mac-distribution/internal/fileutil/copy.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"io"
+	"os"
+
+	"github.com/gravitational/trace"
+)
+
+// CopyFile copies a file from src to dst.
+func CopyFile(src, dst string) error {
+	w, err := os.OpenFile(dst, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0755) // create or overwrite
```
zmb3, line 12: Should we be preserving the permissions of the original file rather than using 0755?

tools/mac-distribution/internal/fileutil/copy.go
```diff
@@ -0,0 +1,29 @@
...
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	defer w.Close()
+
+	r, err := os.OpenFile(src, os.O_RDONLY, 0)
+	if err != nil {
+		return trace.Wrap(err)
```
zmb3, line 20: On error should we be deleting the dst file or leaving it behind?

tools/mac-distribution/internal/zipper/zipper.go
```diff
@@ -0,0 +1,63 @@
...
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/gravitational/trace"
+)
+
+type DirZipper interface {
```
zmb3, line 14: Do we need an interface here? Looks like there's only one implementation of it..

tools/mac-distribution/internal/zipper/zipper.go
```diff
@@ -0,0 +1,63 @@
...
+	root := filepath.Clean(dir)
+
+	err := filepath.WalkDir(root, fs.WalkDirFunc(func(path string, d fs.DirEntry, err error) error {
+		// Ignore zipping directories
+		if d.IsDir() {
+			return nil
+		}
+
+		if !opts.IncludePrefix {
+			path, _ = strings.CutPrefix(path, root+string(os.PathSeparator))
```
zmb3, line 45: Is `TrimPrefix` more appropriate here?

tools/mac-distribution/notarize/submit.go
```diff
@@ -0,0 +1,81 @@
...
+	args := []string{
+		"notarytool",
+		"submit", pathToPackage,
+		"--team-id", t.Creds.BundleID,
+		"--apple-id", t.Creds.AppleUsername,
+		"--apple-password", t.Creds.ApplePassword,
+		"--output-format", "json",
+	}
+
+	stdout, err := t.cmdRunner.RunCommand("xcrun", args...)
```
r0mant, line 41: Like mentioned above, in dry run mode this looks like it will dump all arguments, including password, into the logs.

tools/mac-distribution/notarize/submit.go
```diff
@@ -0,0 +1,81 @@
...
+func (t *Tool) WaitForSubmission(id string) error {
+	args := []string{
+		"notarytool",
+		"wait", id,
+		"--team-id", t.Creds.BundleID,
+		"--apple-id", t.Creds.AppleUsername,
+		"--apple-password", t.Creds.ApplePassword,
+		"--output-format", "json",
+	}
+	stdout, err := t.cmdRunner.RunCommand("xcrun", args...)
```
r0mant, line 74: Same as above regarding arguments logging.

tools/mac-distribution/packaging/appbundle.go
```diff
@@ -0,0 +1,112 @@
...
+	return &AppBundlePackager{
+		Info:       info,
+		log:        log,
+		notaryTool: opts.NotaryTool,
+	}
+}
+
+// Package creates an app bundle from the provided info.
+func (a *AppBundlePackager) Package() error {
+	if err := a.Info.validate(); err != nil {
```
r0mant, line 56: Why are we validating this inside "Package()" rather than inside the NewAppBundlePackager "constructor"?

tools/mac-distribution/packaging/appbundle.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	// Copy the app binary into the skeleton
+	binDir := filepath.Join(a.Info.Skeleton, "Contents", "MacOS")
+	if err := os.MkdirAll(binDir, 0755); err != nil {
+		return trace.Wrap(err)
+	}
+
+	binDest := filepath.Join(binDir, filepath.Base(a.Info.AppBinary))
+	if err := fileutil.CopyFile(a.Info.AppBinary, binDest); err != nil {
```
camscale, line 67: Should all this be cleaned up when done? Or at least if there's an error?

tools/mac-distribution/packaging/appbundle.go
camscale: Since the `gravitational/trace` library is being used in this module, I'd suggest this should be `trace.Errorf()` here and all places that errors are created other than wrapped.

tools/mac-distribution/packaging/packageinstaller.go
```diff
@@ -0,0 +1,169 @@
...
+	Logger     *slog.Logger
+	DryRun     bool
+}
+
+// NewPackageInstallerPackager creates a new PackageInstallerPackager.
+func NewPackageInstallerPackager(info *PackageInstallerInfo, opts *PackageInstallerPackagerOpts) *PackageInstallerPackager {
+	log := opts.Logger
+	if log == nil {
+		log = slog.Default()
+	}
```
camscale, line 54: same comment re: `cmp.Or()`

tools/mac-distribution/main.go
doggydogworld: Just ended up using `trace.BadParameter`. I used that in the other CLI's so this should be consistent.

tools/mac-distribution/internal/exec/exec.go
doggydogworld: Updated this to use `[]byte`. Was maybe like 2 lines of changes.

tools/mac-distribution/internal/fileutil/copy.go
doggydogworld: Yeah I had forgotten to add that in. Original implementation was for the app bundle binary which might need to be 0755. This now preserves by default but allows for overriding permissions.

tools/mac-distribution/internal/zipper/zipper.go
```diff
@@ -0,0 +1,63 @@
...
+	return &DefaultZipper{}
+}
+
+// DefaultZipper is the default implementation of DirZipper
+type DefaultZipper struct{}
+
+// ZipDir will zip the directory into the specified output file
+func (z *DefaultZipper) ZipDir(dir string, out io.Writer, opts DirZipperOpts) error {
+	zipwriter := zip.NewWriter(out)
+	defer zipwriter.Close()
```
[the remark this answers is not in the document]
doggydogworld: Implemented this. Forgot to add errors checking for a lot of the `Close()`

tools/mac-distribution/packaging/appbundle.go
doggydogworld: Added both of these changes.

tools/mac-distribution/go.mod
```diff
@@ -0,0 +1,16 @@
+module github.com/gravitational/shared-workflows/tools/mac-distribution
+
+go 1.23.2
```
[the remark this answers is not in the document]
doggydogworld: Looks like 1.24.0 was just released today!

tools/mac-distribution/internal/fileutil/copy.go
doggydogworld: Updated to attempt deletion on encountering an error.

tools/mac-distribution/go.mod
```diff
@@ -0,0 +1,16 @@
+module github.com/gravitational/shared-workflows/tools/mac-distribution
+
+go 1.24.0
+
+require (
+	github.com/alecthomas/kong v1.8.1
+	github.com/gravitational/trace v1.5.0
```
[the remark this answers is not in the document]
doggydogworld: Yeah I've seen some talk about this before. It seems we would rather want to reduce dependency on this library. I'll remove the dependency and prefer idiomatic Go error handling.

tools/mac-distribution/README.md
```diff
@@ -0,0 +1,42 @@
+# mac-packaging
+
+## Usage
+
+### Packaging
+
+App Bundle (.app)
+```shell
+mac-distribution package-app tsh tsh.app/
```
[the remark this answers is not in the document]
doggydogworld: I considered this a bit. I think for now keeping it as a standalone build tool keeps the scope simple for now. I'm worried that decisions made in this tool could become conventions made for a single build tool. I'd rather let 2 or 3 build tools be created and then consolidate differences later to determine what makes sense. The way the `kong` library handles modularization of CLI commands should make it easy to move things into a single build tool later.

I am a bit torn because I think a single consolidated build tool will eventually be what we want.

tools/mac-distribution/main.go
```diff
@@ -0,0 +1,168 @@
...
+		notarize.Creds{
+			AppleUsername:   g.AppleUsername,
+			ApplePassword:   g.ApplePassword,
+			SigningIdentity: g.SigningID,
+			TeamID:          g.TeamID,
+		},
+		append(
+			extraOpts,
+			notarize.MaxRetries(g.Retry),
+			notarize.WithLogger(log),
```
camscale, line 157: Can these be set when initializing `extraOpts`?

    extraOpts := []notarize.Opt{notarize.MaxRetries(g.Retry), notarize.WithLogger(log)}

That would simplify this `NewTool()` expression somewhat,

camscale (approved):

tools/mac-distribution/README.md
r0mant, line 27: Did you disable CLI flags?

tools/mac-distribution/notarize/submit.go
```diff
@@ -0,0 +1,141 @@
...
+// Submit will submit a package for notarization.
+// A package must be in the format of a zip, package installer (.pkg), or a disk image (.dmg)
+// A success will return a submission ID which can be polled on later.
+func (t *Tool) Submit(pathToPackage string) (id string, err error) {
+	args := []string{
+		"notarytool",
+		"submit", pathToPackage,
+		"--team-id", t.Creds.TeamID,
+		"--apple-id", t.Creds.AppleUsername,
+		"--password", t.Creds.ApplePassword,
```
r0mant, line 56: Did you make sure that the command's arguments do not get written to any of the outputs/logs? Can it maybe take the credentials from environment?

Same for wait command below.

r0mant (commented):
Couple of comments before I approve.

PR #330, opened 2025-02-11
Support decoding `--reviewers` as plain-text JSON

fheinecke:
The shared-workflows bot currently only supports decoding the `--reviewers` flag as a base64-encoded JSON string. For example, rather than

```json
{
  "reviewer-group-1": {
    "username": {
        ... // Repeat for 100+ lines
```

the tool requires the value as `ewogICJyZXZpZXdlci1ncm91cC0xIjogewogICAgInVzZXJuYW1lIjogewogICAgICAgIC4uLiAvLyBSZXBlYXQgZm9yIDEwMCsgbGluZXMK...`. For those of us who cannot base64-decode a string in their head, this is really difficult to read and debug.

Previously this was needed because GHA environment variables/secrets do not "play nice" with structured multiline data. However, now that we're starting to use [`env-loader`](https://github.com/gravitational/shared-workflows/tree/main/tools/env-loader) for these values, we can support plain-text JSON without issue.

This change is fully backwards compatible with all current uses of the bot.

### DIFF
--- a/bot/main.go
+++ b/bot/main.go
@@ -149,7 +149,7 @@ func parseFlags() (flags, error) {
 
 	var decodedReviewers string
 	// Support base64-encoded JSON reviewer string and just JSON without base64 encoding
-	if reviewers != nil && strings.Contains(*reviewers, "{") {
+	if strings.Contains(*reviewers, "{") {
 		decodedReviewers = *reviewers
 	} else {
 		reviewerBytes, err := base64.StdEncoding.DecodeString(*reviewers)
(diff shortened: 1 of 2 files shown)


doggydogworld (approved):

bot/main.go
```diff
@@ -147,9 +147,16 @@ func parseFlags() (flags, error) {
 		return flags{}, trace.BadParameter("reviewers required for assign and check")
 	}
 
-	data, err := base64.StdEncoding.DecodeString(*reviewers)
-	if err != nil {
-		return flags{}, trace.Wrap(err)
+	var decodedReviewers string
+	// Support base64-encoded JSON reviewer string and just JSON without base64 encoding
+	if strings.Contains(*reviewers, "{") {
```
r0mant, line 152: Nit: Should we make this check a little more strict and use "starts with {" (i.e. use HasPrefix instead of Contains)?

r0mant (approved):

PR #331, opened 2025-02-14
Implementing GitHub webhook handler

doggydogworld:
Created an implementation for the GitHub webhook. It is just a small wrapper around the go-github library.

Made some changes to the main function to use `errgroup` to run the `eventSources` concurrently. The `errgroup` library handles propagating cancellations that result from errors and is similar in functionality to waitgroups. In the event of an error from one of the `eventSources`, the `errgroup` will cancel the context which all goroutines in the group can react to appropriately.

### DIFF
--- a/tools/approval-service/cmd/approval-service.go
+++ b/tools/approval-service/cmd/approval-service.go
@@ -1,5 +1,19 @@
 package main
 
+import (
+	"context"
+	"log"
+	"log/slog"
+	"net/http"
+
+	"github.com/google/go-github/v69/github"
+	"github.com/gravitational/shared-workflows/libs/github/webhook"
+	"github.com/gravitational/trace"
+	"golang.org/x/sync/errgroup"
+)
+
+var logger = slog.Default()
+
 // Process:
 // 1. Take in events from CI/CD systems
 // 2. Extract common information
@@ -28,19 +42,22 @@ func main() {
 	eventSources := []EventSource{
 		NewGitHubEventSource(processor),
 	}
-
 	for _, eventSource := range eventSources {
 		_ = eventSource.Setup()
 	}
 
-	done := make(chan struct{}) // TODO replace with error?
+	// 3. Start event sources
+	eg, ctx := errgroup.WithContext(context.Background())
 	for _, eventSource := range eventSources {
-		_ = eventSource.Run(done)
+		eg.Go(func() error {
+			return eventSource.Run(ctx)
+		})
 	}
 
 	// Block until an event source has a fatal error
-	<-done
-	close(done)
+	if err := eg.Wait(); err != nil {
+		log.Fatal(err)
+	}
 }
 
 // This contains information needed to process a request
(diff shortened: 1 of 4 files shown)


tools/approval-service/cmd/approval-service.go
```diff
@@ -114,43 +134,73 @@ func NewGitHubEventSource(processor ApprovalProcessor) *GitHubEventSource {
...
+		case *github.DeploymentReviewEvent:
+			deployReviewChan <- event
+			return nil
+		default:
+			return trace.Errorf("unknown event type: %T", event)
+		}
+	})
+	mux.Handle("/webhook", webhook.NewHandler(
+		eventProcessor,
+		webhook.WithSecretToken([]byte("secret-token")), // TODO: get from config
```
fheinecke, line 152: Maybe this should always take a string? It's a string when configured with GH, and I don't see when we would want to pass a byte array to this func.

tools/approval-service/cmd/approval-service.go
```diff
@@ -114,43 +134,73 @@ func NewGitHubEventSource(processor ApprovalProcessor) *GitHubEventSource {
...
-// Listen for incoming webhook events. Register HTTP routes, start server, etc. Long running, blocking.
-func (ghes *GitHubEventSource) listenForPayloads(payloads chan interface{}, done chan struct{}) {
-	// Once a call is received, it should return a 200 response immediately.
+	var err error
+	// This will block until an error occurs or the context is done
+	select {
+	case err = <-errc:
+		ghes.srv.Shutdown(context.Background()) // Ignore error - we're already handling one
+	case <-ctx.Done():
+		err = ghes.srv.Shutdown(context.Background())
```
fheinecke, line 196: These should probably take a time-limited shutdown context stemming from `context.Background()`.

tools/approval-service/cmd/approval-service.go
```diff
@@ -114,43 +134,73 @@ func NewGitHubEventSource(processor ApprovalProcessor) *GitHubEventSource {
...
+	ghes.deployReviewChan = deployReviewChan
+
+	mux := http.NewServeMux()
+	eventProcessor := webhook.EventHandlerFunc(func(event interface{}) error {
+		switch event := event.(type) {
+		case *github.DeploymentReviewEvent:
+			deployReviewChan <- event
+			return nil
+		default:
+			return trace.Errorf("unknown event type: %T", event)
```
fheinecke, line 147: [This library only exists for Teleport error handling](https://gravitational.slack.com/archives/C08HF8E9F/p1738272989972819?thread_ts=1738188479.237239&cid=C08HF8E9F), so we probably shouldn't be using it anymore outside of the product (here and elsewhere)

fheinecke (commented):
Overall LGTM. We might want to merge the base of this PR into master though, or its PR will become massive by the time we're done.

@r0mant, would you rather we file a bunch of smaller broken/development/not building PRs (which you'll need to review as only eng leads are codeowners), or one massive one later?

PR #333, opened 2025-02-18
[bot] Allow for non-PR events in docpaths

ptgott:
The workflow currently fails if the event is not a pull request, since it attempts to list files for a PR with number 0 and receives a 404 from the GitHub API. Edit the workflow so it does not return an error if the event is not a PR.


ptgott:
Example of a run that fails because of a 404 due to a non-PR event: https://github.com/gravitational/teleport/actions/runs/13394644883/job/37410507330

fheinecke (approved):

doggydogworld (approved):

zmb3 (approved):

PR #338, opened 2025-03-10
Approval Service MVP

doggydogworld:
MVP of the approval service.

The current implementation supports:
* Listens for GitHub Webhooks and creates Teleport Access Requests from them
* Responds to approvals and rejections of Access Requests and updates the GitHub workflow accordingly
* Handling events and updates for multiple GitHub repositories
* Leader election to avoid duplicating Access Requests

Currently missing:
* Audit events not implemented
* Metrics not implemented
* Reconciliation not implemented (e.g. GitHub Deploy Event with no corresponding Access Request)


tools/approval-service/internal/approvalservice/eventprocessor.go
```diff
@@ -0,0 +1,169 @@
...
+	req.SetStaticLabels(map[string]string{
+		"workflow_run_id": strconv.Itoa(int(e.WorkflowID)),
+		"organization":    e.Organization,
+		"repository":      e.Repository,
+		"environment":     e.Environment,
+	})
+	created, err := p.teleportClient.CreateAccessRequestV2(context.TODO(), req)
+	if err != nil {
+		return fmt.Errorf("creating access request %q: %w", name, err)
+	}
```
fheinecke, line 108: This will not handle multiple jobs stemming from the same root request very well. This needs to:

1. Get an unique identifier for the root triggering workflow. For example, `fmt.Printf("%s-%s-%s-%d", e.Organization, e.Repository, e.Environment, e.WorkflowsID)`. If this is too long (it probably can become too long, as we're assuming an attacker can control the environment name) then it should be shortened with a hashing function. This must be deterministic.
2. Check if an access request with this name already exists. If it does and has not yet been approved or denied, do nothing. If it has been approved or denied, notify GH.
3. If a request does not exist, file a new one.

Because multiple routines can enter this at once and attempt to create an access request at the same time, this is a critical section.

tools/approval-service/internal/approvalservice/eventprocessor.go
```diff
@@ -0,0 +1,169 @@
...
+				return fmt.Errorf("getting environment %q: %w", env, err)
+			}
+			p.envNameToID[envEntry(v.Org, v.Repo, env.Name)] = env.ID
+		}
+	}
+
+	return nil
+}
+
+func (p *processor) ProcessDeploymentReviewEvent(e githubevents.DeploymentReviewEvent, valid bool) error {
```
fheinecke, line 82: This should take a context input

tools/approval-service/internal/approvalservice/eventprocessor.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	req.SetRequestReason("Deployment review for environment " + e.Environment)
+	req.SetStaticLabels(map[string]string{
+		"workflow_run_id": strconv.Itoa(int(e.WorkflowID)),
+		"organization":    e.Organization,
+		"repository":      e.Repository,
+		"environment":     e.Environment,
+	})
+	created, err := p.teleportClient.CreateAccessRequestV2(context.TODO(), req)
```
fheinecke, line 105: This context should be time limited. If Teleport hangs, this should not block indefinitely

tools/approval-service/internal/approvalservice/eventprocessor.go
```diff
@@ -0,0 +1,169 @@
...
+		return fmt.Errorf("creating access request %q: %w", name, err)
+	}
+	p.log.Info("Created access request", "access_request_name", created.GetName())
+	return nil
+}
+
+func (p *processor) HandleReview(ctx context.Context, req types.AccessRequest) error {
+	p.log.Info("Handling review", "access_request_name", req.GetName())
+
+	runIDLabel := req.GetStaticLabels()["workflow_run_id"]
```
fheinecke, line 120: All of these need checks to ensure that the map contains the values in question

fheinecke, line 120: Key names should be made `const`

tools/approval-service/internal/approvalservice/eventprocessor.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return fmt.Errorf("generating access request: %w", err)
+	}
+
+	req.SetRequestReason("Deployment review for environment " + e.Environment)
+	req.SetStaticLabels(map[string]string{
+		"workflow_run_id": strconv.Itoa(int(e.WorkflowID)),
+		"organization":    e.Organization,
+		"repository":      e.Repository,
+		"environment":     e.Environment,
```
fheinecke, line 103: This can be controlled by an attacker, making the label value arbitrary length

tools/approval-service/internal/approvalservice/eventprocessor.go
```diff
@@ -0,0 +1,169 @@
...
+	// Create access request
+	// The name of an access request must be a UUID
+	name := uuid.New().String()
+	p.log.Info("Creating access request", "access_request_name", name)
+	req, err := types.NewAccessRequest(name, p.TeleportUser, p.TeleportRole)
+	if err != nil {
+		return fmt.Errorf("generating access request: %w", err)
+	}
+
+	req.SetRequestReason("Deployment review for environment " + e.Environment)
```
fheinecke, line 98: `e.Environment` is assumed to be attacker-controlled and can be an arbitrary length, which Teleport does not handle well. There needs to be a check that fails the request if the value is above a certain length. 255 might be a reasonable starting point.

fheinecke, line 98: This will need workflow call parameters as well, which means accepting another webhook, and might require waiting for the webhook call before filing and access request. Alternatively, if the access request can be modified, the workflow call parameters could update the request when the webhook call is made.

tools/approval-service/internal/approvalservice/eventprocessor.go
```diff
@@ -0,0 +1,169 @@
...
+	if !ok {
+		return fmt.Errorf("encountered unkown environment %q, this indicates a problem during setup", envEntry(orgLabel, repoLabel, env))
+	}
+
+	runID, err := strconv.Atoi(runIDLabel)
+	if err != nil {
+		return fmt.Errorf("parsing workflow run ID: %w", err)
+	}
+
+	state := github.PendingDeploymentApprovalState_APPROVED
```
fheinecke, line 151: In the off chance that more states are added in the future, this should default to rejected, with the state set to "approved" if the request is approved. Alternatively this could implement a `switch` statement with a default case of denial.

tools/approval-service/internal/approvalservice/eventprocessor.go
```diff
@@ -0,0 +1,169 @@
...
+		state = github.PendingDeploymentApprovalState_REJECTED
+	}
+
+	p.githubClient.UpdatePendingDeployment(ctx, github.PendingDeploymentInfo{
+		Org:     orgLabel,
+		Repo:    repoLabel,
+		RunID:   int64(runID),
+		State:   state,
+		EnvIDs:  []int64{envID},
+		Comment: "Approved by pipeline approval service",
```
fheinecke, line 162: Can you update this to include a link or reference to the access request?

tools/approval-service/internal/approvalservice/githubevents/githubevents.go
```diff
@@ -0,0 +1,231 @@
...
+		validation:  map[string]struct{}{},
+	}
+
+	for _, o := range defaultOpts {
+		o(s)
+	}
+
+	for _, o := range opt {
+		o(s)
+	}
```
fheinecke, line 86: nit:
```suggestion
	for _, opt := range append(defaultOpts, opts) {
		opt(s)
	}
```

tools/approval-service/internal/approvalservice/githubevents/githubevents.go
```diff
@@ -0,0 +1,231 @@
...
+type Source struct {
+	processor   DeploymentReviewEventProcessor
+	addr        string
+	secretToken string
+
+	deployReviewChan chan *github.DeploymentReviewEvent
+	listener         net.Listener
+	srv              *http.Server
+
+	validation map[string]struct{}
```
fheinecke, line 26: Is this just a placeholder for actual validation functions?

tools/approval-service/internal/approvalservice/githubevents/githubevents.go
```diff
@@ -0,0 +1,231 @@
...
+	// Other fields TODO. Potential fields:
+	// * Source
+	// * Commit/tag/source control identifier
+	// * "Parameters" map. In the case of GHA, this would be
+	//   any input provided to a workflow dispatch event
+	// See RFD for more details
+}
+
+// Opt is a functional option for the GitHub event source.
+type Opt func(*Source)
```
fheinecke, line 59:
```suggestion
type SourceOpt func(*Source)
```

tools/approval-service/internal/approvalservice/githubevents/githubevents.go
```diff
@@ -0,0 +1,231 @@
...
+		s.log = logger
+	}
+}
+
+var defaultOpts = []Opt{
+	WithLogger(slog.Default()),
+}
+
+func NewSource(cfg config.GitHubEvents, processor DeploymentReviewEventProcessor, opt ...Opt) *Source {
+	s := &Source{
```
fheinecke, line 73: nit: consistent naming with receiver funcs
```suggestion
	ghes := &Source{
```

tools/approval-service/internal/approvalservice/githubevents/githubevents.go
```diff
@@ -0,0 +1,231 @@
...
+		}
+	}
+
+	return s
+}
+
+// Setup GH client, webhook secret, etc.
+// https://github.com/go-playground/webhooks may help here
+func (ghes *Source) Setup() error {
+	deployReviewChan := make(chan *github.DeploymentReviewEvent)
```
fheinecke, line 100: This will block new events until the first received event has been processed. Capacity should probably be set to some reasonably large number (128?)

tools/approval-service/internal/approvalservice/githubevents/githubevents.go
```diff
@@ -0,0 +1,231 @@
...
+		return fmt.Errorf("error performing automated checks: %w", err)
+	}
+
+	event := DeploymentReviewEvent{
+		Requester:    payload.GetRequester().GetLogin(),
+		Environment:  payload.GetEnvironment(),
+		Organization: payload.GetOrganization().GetLogin(),
+		Repository:   payload.GetRepo().GetName(),
+		WorkflowID:   payload.GetWorkflowRun().GetWorkflowID(),
+	}
```
fheinecke, line 192: There needs to be some check to see if these are actually set

tools/approval-service/internal/approvalservice/githubevents/githubevents_test.go
```diff
@@ -0,0 +1,71 @@
...
+			require.NoError(t, err)
+			if d.IsDir() {
+				return nil
+			}
+
+			payloadFile, err := os.Open(path)
+			require.NoError(t, err)
+			defer payloadFile.Close()
+
+			req, err := http.NewRequest("POST", "http://"+webhook.getAddr()+"/webhook", payloadFile)
```
fheinecke, line 49: This should be built with `net/url`

tools/approval-service/internal/approvalservice/githubevents/githubevents.go
```diff
@@ -0,0 +1,231 @@
...
+			return fmt.Errorf("unknown event type: %T", event)
+		}
+	})
+	mux.Handle("/webhook", webhook.NewHandler(
+		eventProcessor,
+		webhook.WithSecretToken(ghes.secretToken),
+		webhook.WithLogger(ghes.log),
+	))
+
+	ln, err := net.Listen("tcp", ghes.addr)
```
fheinecke, line 119:
```suggestion
	listener, err := net.Listen("tcp", ghes.addr)
```

tools/approval-service/internal/approvalservice/githubevents/githubevents.go
```diff
@@ -0,0 +1,231 @@
...
+	var err error
+	// This will block until an error occurs or the context is done
+	select {
+	case err = <-errc:
+		ghes.srv.Shutdown(context.Background()) // Ignore error - we're already handling one
+	case <-ctx.Done():
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		err = ghes.srv.Shutdown(ctx)
+		<-errc // flush the error channel to avoid a goroutine leak
```
fheinecke, line 167: Should this be captured/aggregated instead?

tools/approval-service/tool/webhooktest/main.go
```diff
@@ -0,0 +1,33 @@
+package main
+
+import (
+	"embed"
+	"log"
+	"net/http"
+)
+
+//go:embed payload.json
+var payloadFile embed.FS
```
fheinecke, line 10: Why does this need to be embedded?

fheinecke (commented):
`make lint` reports an error as well - please take a look.

I talked with Roman and we thinks that using dynamodb for storing state should be reconsidered. This would solve a couple of problems:
* It keeps arbitrary length data out of Teleport.
* We can use it for audit logging. The application can be configured to use it as an append-only ledger, and cloudtrail/panther should keep a record of all change requests from all identities should we think that infra was compromised.
* We're probably going to end up needing to store a lot more state information (workflow runs that trigger while an approval is pending, workflow dispatch call inputs, etc.)

I wouldn't be opposed to using pgaudit with aurora postgres either, and S3 recently added a database service which might fit our needs.

What do you think about using a database for state storage?

libs/github/webhook/webhook.go
```diff
@@ -0,0 +1,163 @@
...
+type EventHandlerFunc func(event interface{}) error
+
+// Handler is an implementation of [http.Handler] that handles GitHub webhook events.
+type Handler struct {
+	eventHandler EventHandlerFunc
+	secretToken  []byte
+	log          *slog.Logger
+}
+
+var _ http.Handler = &Handler{}
```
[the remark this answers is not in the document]
doggydogworld: This is just an assertion that it satisfies the `http.Handler` interface. Kind of a way to "document" it.

tools/approval-service/internal/approvalservice/eventprocessor.go
```diff
@@ -0,0 +1,169 @@
...
+	TeleportRole string
+
+	teleportClient teleClient
+	githubClient   ghClient
+	log            *slog.Logger
+
+	// envNameToID is a cache of environment names to their IDs.
+	// This is expected to be read-only after Setup() is called.
+	// If this is not the case, it will need to be protected by a mutex.
+	envNameToID map[string]int64
```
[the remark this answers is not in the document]
doggydogworld: Hmm from their [docs](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment#deleting-an-environment) it seems you need to be an admin to delete an environment. While the attack could work assuming non-uniqueness of IDs, an admin would be able to bypass these checks anyway. The secrets and variables under it are fair game though for anyone in our org it seems.

It does bring up an interesting point in that a recreated environment would effectively make this data out of date which would require a restart. I couldn't find a documented event for environment CRUD operations but I could add an expiry to entries in the cache.

tools/approval-service/cmd/approval-service.go
```diff
@@ -0,0 +1,103 @@
...
+
+	go func() {
+		<-sigs
+		log.Println("Caught signal, shutting down...")
+		cancel()
+	}()
+
+	select {
+	case err := <-errc:
+		log.Fatal(err)
```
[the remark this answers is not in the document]
doggydogworld: I ended up simplifying this a bit. A cancelled context will result in `Run(context.Context) error` returning an error no matter what so we only need to listen on the channel it sends to.

tools/approval-service/internal/approvalservice/eventprocessor.go
doggydogworld: For now I added the Access Request ID. The object doesn't seem to have a link directly to it. We could probably construct it later though from the configured URL.

tools/approval-service/tool/webhooktest/main.go
doggydogworld: Removed this file as I just used it for initial testing.

r0mant (commented):
@doggydogworld This PR is huge, can you split it up in more reasonable parts that build on top of each other and can be reviewed independently. For example:
- PR 1: Implementation of the service itself.
- PR 2: Helm chart for the deployment
- PR 3: IaC for Teleport and other related resources

doggydogworld:
> @doggydogworld This PR is huge, can you split it up in more reasonable parts that build on top of each other and can be reviewed independently. For example:
> 
>     * PR 1: Implementation of the service itself
>       
>       * _Maybe even this should be split in a series of PRs_
> 
>     * PR 2: Helm chart for the deployment
> 
>     * PR 3: IaC for Teleport and other related resources

Yeah I was looking through it and my computer was lagging from the size haha. Will split it out.

The implementation of the service itself shouldn't be too big. A good amount of the LoC currently is IaC and some tests.

PR #339, opened 2025-03-12
Update dependency-review action

zmb3:
We want at least v4.3.3 in order to get the fix from https://github.com/actions/dependency-review-action/pull/767 that prevents large summaries from failing the job.


zmb3:
Looks like this is going to be a more disruptive upgrade: https://github.com/actions/dependency-review-action/issues/809#issuecomment-2285930024

Maybe we should just disable summaries for now.

PR #347, opened 2025-03-17
gha(path-filter) fix quotes

taraspos:
Add quotes around `*` to fix failures like:
```
Error: unidentified alias "*/*.csv" (2:13)

 1 | csv:
 2 |   - **/*.csv
-----------------^
```

PR #348, opened 2025-03-17
gha(path-filter) add permissions to read PR

taraspos:
Followup for the:
- https://github.com/gravitational/shared-workflows/pull/347
- https://github.com/gravitational/shared-workflows/pull/344


taraspos:
Seems like `pull-requests: read` permission might be needed on the parent job as well

PR #360, opened 2025-04-08
Attempt to fix dependency review workflow

fheinecke:
Merge queue checks are failing because this GH-managed action is producing a step summary that is large enough that GitHub rejects it. This _attempts_ to work around the issue by getting the action to write to a different path, and then conditionally uploading the output to the step summary if it is small enough.

Upstream issue: https://github.com/actions/dependency-review-action/issues/867


fheinecke:
This approach does not work and there does not seem to be any way to make this approach work

PR #363, opened 2025-04-15
Update CSV linter to ignore removed files

r0mant:
Currently the action considers removed CSV files as "changed" too which results in the action failing when trying to lint a removed file. This change should make the action only consider added or modified files. The syntax used is documented here: https://github.com/dorny/paths-filter?tab=readme-ov-file#advanced-options.

### DIFF
--- a/.github/workflows/csv-lint.yaml
+++ b/.github/workflows/csv-lint.yaml
@@ -25,7 +25,7 @@ jobs:
           list-files: shell
           filters: |
             csv:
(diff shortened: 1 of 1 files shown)


r0mant:
The change added here worked on the PR where it was previously failing due to deleted file so this should be good to go: https://github.com/gravitational/cloud-terraform/pull/6379.

russjones (approved):

PR #367, opened 2025-05-03
Approval Service MVP - Processors

doggydogworld:
## Overview

This is the second part of the Approval Service MVP ([previous](https://github.com/gravitational/shared-workflows/pull/366)]. Contains the following:
* GitHub Processor responsible for processing events related to GitHub deployments. It handles creating Access Requests and also handling the approval/rejection of access requests
* Custom implementation for GitHub App Client. There are libraries out there for this but in an effort to keep our dependency footprint smaller I just implemented it myself.


r0mant (commented):
@doggydogworld Can this one be split up further? For example, move out the reconciler into a separate PR, and possibly anything else that you think might make sense to review independently.

doggydogworld:
> @doggydogworld Can this one be split up further? For example, move out the reconciler into a separate PR, and possibly anything else that you think might make sense to review independently.

I split the PR up so this one is more focused.

tools/approval-service/internal/approvalservice/eventprocessor/eventprocessor.go
```diff
@@ -0,0 +1,107 @@
...
+
+// Run starts the Processor and starts receiving events from the event sources.
+func (p *Processor) Run(ctx context.Context) error {
+	eg, ctx := errgroup.WithContext(ctx)
+
+	eg.Go(func() error {
+		return p.githubEventListener(ctx)
+	})
+
+	return eg.Wait()
```
camscale, line 92: I don't see what use an error group is here. Can't this just be:

    return p.gitHubEventListener(ctx)

Or is this going to turn into a bunch more stuff here soon?

tools/approval-service/internal/approvalservice/eventprocessor/githubprocessor/githubhandler.go
```diff
@@ -0,0 +1,249 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("opening private key: %w", err)
+	}
+	defer func() {
+		err = errors.Join(err, f.Close())
+	}()
+	pKey, err := io.ReadAll(f)
+	if err != nil {
+		return nil, fmt.Errorf("reading private key: %w", err)
+	}
```
camscale, line 96: [`os.ReadFile()`](https://pkg.go.dev/os#ReadFile) would be simpler

tools/approval-service/internal/approvalservice/eventprocessor/githubprocessor/githubhandler.go
```diff
@@ -0,0 +1,249 @@
...
+func (p *Processor) FindExistingAccessRequest(ctx context.Context, e githubevents.DeploymentReviewEvent) (types.AccessRequest, error) {
+	list, err := p.teleClient.GetAccessRequests(ctx, types.AccessRequestFilter{})
+	if err != nil {
+		return nil, fmt.Errorf("getting access requests: %w", err)
+	}
+	for _, req := range list {
+		labels := req.GetStaticLabels()
+		if labels == nil {
+			continue
+		}
```
camscale, line 114: not needed - lookup of a `nil` map is valid

tools/approval-service/internal/approvalservice/eventprocessor/githubprocessor/githubhandler.go
```diff
@@ -0,0 +1,249 @@
...
+			return req, nil
+		}
+	}
+
+	// No existing access request found.
+	return nil, nil
+}
+
+// CreateNewAccessRequest creates a new Access Request for the given GitHub deployment review event.
+func (p *Processor) CreateNewAccessRequest(ctx context.Context, e githubevents.DeploymentReviewEvent, req types.AccessRequest) (types.AccessRequest, error) {
```
camscale, line 125: nit: CreateNew.. is a bit verbose. I expect all things created are new at the time of creation. Perhaps just `CreateAccessRequest`.

tools/approval-service/internal/approvalservice/eventprocessor/githubprocessor/githubhandler.go
```diff
@@ -0,0 +1,249 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("storing data: %w", err)
+	}
+
+	newReq, err := p.teleClient.CreateAccessRequestV2(ctx, req)
+	if err != nil {
+		return nil, fmt.Errorf("creating access request: %w", err)
+	}
+
+	_, denyErr := p.automaticallyDenied(e)
```
camscale, line 152: if we're going to deny an access request, why create it in the first place? is this needed for auditing purposes?

tools/approval-service/internal/approvalservice/eventprocessor/githubprocessor/githubhandler.go
```diff
@@ -0,0 +1,249 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("getting workflow run info: %w", err)
+	}
+
+	reason, err := genReason(e, runInfo)
+	if err != nil {
+		return nil, fmt.Errorf("generating reason template: %w", err)
+	}
+	req.SetRequestReason(reason)
+	err = p.store.StoreWorkflowInfo(ctx, req, store.GitHubWorkflowInfo{
```
camscale, line 137: is it correct to store the workflow info and access request before we are able to successfully create it and not deny it? Should this be moved down after the other parts?

tools/approval-service/internal/approvalservice/eventprocessor/githubprocessor/githubhandler.go
```diff
@@ -0,0 +1,249 @@
...
+	p.log.Info("Updating deployment", "access_request_name", req.GetName(), "environment", data.Env, "state", state)
+	if err := p.client.ReviewDeploymentProtectionRule(ctx, data.Org, data.Repo, data.WorkflowRunID, state, data.Env,
+		fmt.Sprintf("Access by pipeline approval service. Access Request ID: %s", req.GetName())); err != nil {
+		return fmt.Errorf("reviewing deployment protection rule: %w", err)
+	}
+
+	return nil
+}
+
+var reasonTmpl = template.Must(template.New("reason").Parse(`GitHub Deployment Review for:
```
camscale, line 210: These singleton templates do not need a name to `New()` - I always just pass an empty string as that name has no meaning.

libs/github/github.go
```diff
@@ -52,6 +57,159 @@ func New(ctx context.Context, token string) (*Client, error) {
...
+	}
+
+	return tr, nil
+}
+
+// RoundTrip implements the http.RoundTripper interface for the jwtTransport.
+func (j *jwtAuthTransport) RoundTrip(orig *http.Request) (*http.Response, error) {
+	if orig.Body != nil {
+		// As per the http.RoundTripper contract, we should close the body after we're done with it.
+		defer orig.Body.Close()
```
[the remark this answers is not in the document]
doggydogworld: Did a quick check on the `go-github` library for reference and they don't bother closing the body. Removed this logic from the code.

libs/github/github.go
```diff
@@ -52,6 +57,159 @@ func New(ctx context.Context, token string) (*Client, error) {
...
+// For more details: https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-an-installation-access-token-for-a-github-app
+type installationAuthTransport struct {
+	// tr is the underlying http.RoundTripper that will be used to make requests.
+	tr http.RoundTripper
+
+	appsClient     *go_github.Client // GitHub client for making API requests to the GitHub Apps API
+	installationID int64
+
+	token *go_github.InstallationToken
+	mu    sync.Mutex // mutex to protect access to the token
```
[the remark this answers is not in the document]
doggydogworld: Yeah my brain ended up flipping this around! I guess calling it a "mutex hat" is a good way to remember which side it goes on haha.

tools/approval-service/internal/approvalservice/eventprocessor/store/processor.go
```diff
@@ -0,0 +1,51 @@
...
+type processorService struct {
+}
+
+const procIDLabel = "procid"
+
+func (p *processorService) StoreProcID(ctx context.Context, req types.AccessRequest, procID string) error {
+	labels := req.GetStaticLabels()
+	if labels == nil {
+		labels = make(map[string]string)
+	}
```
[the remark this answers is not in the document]
doggydogworld: Unfortunately it seems like `map[string]string` does not satisfy `comparable` (ref: https://pkg.go.dev/builtin#comparable)

Won't compile with this.

tools/approval-service/internal/approvalservice/eventprocessor/eventprocessor.go
doggydogworld: Meant to add asynchronous handling of Access Requests but forgot. Ended up just simplifying the code a bit.

tools/approval-service/internal/approvalservice/eventprocessor/eventprocessor.go
```diff
@@ -0,0 +1,107 @@
...
+		if logger == nil {
+			return errors.New("logger cannot be nil")
+		}
+		p.log = logger
+		return nil
+	}
+}
+
+// New creates a new Processor instance.
+func New(ctx context.Context, teleConfig config.Teleport, sp *Consumers, coordinator coordination.Coordinator, opts ...Opt) (*Processor, error) {
```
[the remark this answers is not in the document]
doggydogworld: Yeah I was running through multiple different names for this but had a generic placeholder originally. Should be more consistent now.

r0mant (changes requested):
I am having a very hard time building the mental model of how this works. IMO it's significantly over-engineered, we have dispatchers, event processors, leader elections, different stores, repositories, reconcilers, etc., all of this to just be able to get events from Github, create an access request, wait for it approval and then approve the build.

@doggydogworld I want us to _massively_ simplify the architecture of this. Let's get rid of all inconsequential stuff like unneeded interfaces, leader elections (just deploy 1 replica), different store types, reduce the levels of indirection, etc.

tools/approval-service/internal/eventprocessor/workflowevents.go
```diff
@@ -0,0 +1,433 @@
...
+			return
+		}
+
+		// Existing request is already reviewed (approved/rejected).
+		// This happens when a previous job in the workflow has already triggered the review process.
+		_ = w.HandleAccessRequestReviewed(ctx, existingRequest) // Async handling does not return an error.
+		return
+	}
+
+	req, err := w.createAccessRequest(ctx, e)
```
r0mant, line 181: What happens if this (or anything else in the core event handling loop) fails? The deployment is waiting for approval but we failed to create an access request for example. How do we unblock the release?

tools/approval-service/internal/eventprocessor/workflowevents.go
```diff
@@ -0,0 +1,433 @@
...
+	for _, req := range list {
+		info, err := w.store.GetWorkflowInfo(ctx, req)
+		if err != nil {
+			w.log.Debug("failed to get workflow info for access request", "access_request_name", req.GetName(), "error", err)
+			// Not all Access Requests will have workflow info, so we can ignore this error.
+			continue
+		}
+
+		// Check if the Access Request matches the GitHub deployment review event.
+		if info.Org == e.Organization && info.Repo == e.Repository && info.Env == e.Environment && info.WorkflowRunID == e.WorkflowID {
```
r0mant, line 213: Nit: I'd make a helper "IsEqual" method on the object.

tools/approval-service/internal/eventprocessor/workflowevents.go
```diff
@@ -0,0 +1,433 @@
...
+	w.log.Info("Handled access request reviewed", "access_request_name", req.GetName(), "org", info.Org, "repo", info.Repo)
+}
+
+func (w *WorkflowEventsProcessor) getApprovalHandler(org, repo string) (*GitHubDeploymentApprovalHandler, error) {
+	handler, ok := w.deploymentApprovalHandlers[formatGitHubRepoKey(org, repo)]
+	if !ok {
+		return nil, fmt.Errorf("the organization %q and repository %q are not configured for deployment approval", org, repo)
+	}
+	return handler, nil
+}
```
r0mant, line 317: Like I suggested above, this should go away. Let's just make it work for teleport and worry about others later.

tools/approval-service/internal/eventprocessor/workflowevents.go
```diff
@@ -0,0 +1,433 @@
...
+
+// newGitHubDeploymentApprovalHandler creates a new GitHub deployment approval handler for deployment protection rules
+// in a given GitHub organization and repository.
+func newGitHubDeploymentApprovalHandler(ctx context.Context, cfg config.GitHubSource, log *slog.Logger) (*GitHubDeploymentApprovalHandler, error) {
+	key, err := os.ReadFile(cfg.Authentication.App.PrivateKeyPath)
```
r0mant, line 322: Can this be passed to the `ReleaseService` via config rather than having the service read in the file itself?

tools/approval-service/internal/eventprocessor/store/github.go
```diff
@@ -0,0 +1,194 @@
...
+	Env string
+	// WorkflowRunID is the ID of the workflow run that is waiting for approval.
+	WorkflowRunID int64
+}
+
+type GitHubStorerOpt func(g *githubStore) error
+
+type githubStore struct {
+	log *slog.Logger
+}
```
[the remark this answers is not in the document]
doggydogworld: Removed the `store` struct. Now they're just functions.

tools/approval-service/internal/eventprocessor/workflowevents.go
doggydogworld: That will be what the reconciling loop is for (one more PR for that). It will scan for "blocked" workflows and will push an event to the ReleaseService.

It will also handle the failure mode for approved/rejected Access Requests in that it will scan for those and will push those as new events to the ReleaseService as well.

libs/github/deployment_review.go
```diff
@@ -48,3 +46,24 @@ func (c *Client) ReviewDeploymentProtectionRule(ctx context.Context, org, repo s
 
 	return nil
 }
+
+type PendingDeploymentInfo struct {
+	Environment string
+}
```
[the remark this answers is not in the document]
doggydogworld: Added some godocs. This one isn't strictly necessary for this service actually but would be necessary to recover from the failure mode of a "blocked" workflow that hasn't created an Access Request. It's a bit annoying but listing blocked workflow runs doesn't give info about what environment it is requesting which is what this extra call is needed for.

The rest of the information isn't necessary. It's stuff like how long ago did was it requested and the users that can approve it.

camscale:
For some reason, this PR has changes to the go.mod and go.sum of the changelog tool. Why is that? If it is needed, can it be split out? I noticed that the last approvals service PR make changes to the changelog go.mod file too. It is easier to understand the history if these sorts of things are kept separate with a useful description as to why the change is being made.

camscale (approved):
I still feel that there is to much abstraction around the receiving of events, but perhaps that becomes more useful in later PRs? It seems to me that the `ReleaseService` could contain the `http.Handler` for receiving deployment requests, parse it and deal with it there and that would cut out a lot of the input chain theough the various event sources, etc.

I would also not be including the `WorkflowDispatchEvent` stuff in this PR - the implementation just says "unimplemented", so I'd just leave it out and add it all at once when it can be reviewed cohesively. Right now, I have no idea why we need workflow dispatch events so I can't comment on any of that code added here.

libs/github/github.go
```diff
@@ -52,6 +57,155 @@ func New(ctx context.Context, token string) (*Client, error) {
 	}, nil
 }
 
+// NewForApp returns a new GitHub Client with authentication for a GitHub App.
+func NewForApp(ctx context.Context, appID int64, installationID int64, privateKey []byte) (*Client, error) {
+	appTr, err := newAppTransport(ctx, appID, installationID, privateKey)
+	if err != nil {
+		return nil, err
```
r0mant, line 64: Nit: Wrap error.

tools/approval-service/internal/service/githubapproval.go
```diff
@@ -0,0 +1,153 @@
...
+
+// gitHubWorkflowApprover handles approvals/rejections for deployment protection reviews on workflows
+// based on the reviewed Access Requests. This is per-repo, and contains the logic to handle the
+// decision-making process for deployment protection rules.
+type gitHubWorkflowApprover struct {
+	ghClient *github.Client
+	org      string
+	repo     string
+
+	envToRole map[string]string
```
r0mant, line 40: Nit: Explain in a godoc what envToRole is for.

tools/approval-service/internal/service/githubapproval.go
```diff
@@ -0,0 +1,153 @@
...
+
+Repository: {{ .Organization}}/{{ .Repository }}
+Workflow name: {{ .WorkflowName }}
+URL: {{ .URL }}
+Environment: {{ .Environment }}
+Workflow run ID: {{ .WorkflowID }}
+Requester: {{ .Requester }}
+
+This request was generated by the pipeline approval service.
+`))
```
r0mant, line 143: Can we include release version number here or will it be visible in any of the existing fields?

tools/approval-service/internal/service/githublabels.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+const (
+	workflowRunLabel  = "workflow_run_id"
+	organizationLabel = "organization"
+	repositoryLabel   = "repository"
+	environmentLabel  = "environment"
+)
+
+func setWorkflowLabels(req types.AccessRequest, info githubWorkflowLabels) error {
```
r0mant, line 64: Add godocs to set/get labels methods to explain why we're setting labels on the access request.

tools/approval-service/internal/service/releaseservice.go
```diff
@@ -0,0 +1,294 @@
...
+func (w *ReleaseService) HandleDeploymentReviewEventReceived(ctx context.Context, e githubevents.DeploymentReviewEvent) error {
+	w.deploymentReviewEventChan <- e
+	return nil
+}
+
+// HandleWorkflowDispatchEventReceived is a placeholder for processing workflow dispatch events.
+func (w *ReleaseService) HandleWorkflowDispatchEventReceived(ctx context.Context, e githubevents.WorkflowDispatchEvent) error {
+	// This is not implemented yet.
+	return fmt.Errorf("workflow_dispatch event processing is not implemented")
+}
```
r0mant, line 121: Can we just get rid of this completely?

tools/approval-service/internal/service/releaseservice.go
```diff
@@ -0,0 +1,294 @@
...
+	"github.com/gravitational/shared-workflows/tools/approval-service/internal/eventsources/githubevents"
+	teleportclient "github.com/gravitational/teleport/api/client"
+	"github.com/gravitational/teleport/api/types"
+)
+
+// ReleaseService processes GitHub Events and Teleport Access Request state changes that are related to GitHub workflows.
+// At a high level, it waits for a GitHub Workflow to request access to an environment (signaled by deployment_protection_rule),
+// and then creates a Teleport Access Request for the workflow.
+// It then waits for the Access Request to be reviewed, and based on the review, it approves or rejects the workflow in GitHub.
+type ReleaseService struct {
```
r0mant, line 39: It does not appear that the release service business logic has sufficient test coverage. Can you add it?

r0mant (approved):
lgtm once remaining comments are addressed

tools/approval-service/internal/service/githubapproval.go
doggydogworld: The version number is included in the `WorkflowName` since our workflows include the version in the title. I thought about including the tag that was used to build it but I figured it might be redundant in our case.

These are the fields we get to work with though:
* `name`: "Tag Build of gravitational/teleport@v18.1.1"
* `head_branch`: "v18.1.1"

The tag would be more stable if we ever do change our naming.

libs/github/deployment_review.go
```diff
@@ -48,3 +64,27 @@ func (c *Client) ReviewDeploymentProtectionRule(ctx context.Context, org, repo s
 
 	return nil
 }
+
+// PendingDeploymentInfo contains information about a pending deployment that is waiting for approval.
+// It contains environment name and some other metadata that isn't so useful such as how long the deployment has been pending.
```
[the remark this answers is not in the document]
doggydogworld: Just bad wording haha. It's a bit weird but the data the API returns environment name and some other information that isn't that useful. Duration it's been pending and who can approve the request. Fixed the wording so it's more clear

doggydogworld:
> I still feel that there is to much abstraction around the receiving of events, but perhaps that becomes more useful in later PRs? It seems to me that the `ReleaseService` could contain the `http.Handler` for receiving deployment requests, parse it and deal with it there and that would cut out a lot of the input chain theough the various event sources, etc.

The motivation was to abstract the business logic on how the event is sourced and transformed so that there can be extra implementations. I found it would be useful because the information contained in the webhook and the information available from the GitHub API (that the reconciler will use) have similar data but different schemas. I think with the reconciling logic added later it will be more useful.

PR #378, opened 2025-06-20
Approval Service - Coordinator

doggydogworld:
# Overview

Continues: #367

This PR introduces distributed coordination logic. When running more than one instance of the service, each service will receive duplicate events. To avoid duplicating work, we need a way to coordinate work between the services.


doggydogworld:
Keeping to 1 replica for now which will not need distributed coordination.

PR #380, opened 2025-07-01
Fix bug in docpaths workflow

ptgott:
The docpaths workflow detects whether a pull request renames or removes a docs page without introducing adequate redirects, preventing 404s.

If a pull request changes a docs page (e.g.,
`docs/pages/installation.mdx`) to a category index page (e.g., `docs/pages/installation/installation.mdx`), the URL path for the page does not change, and there is no redirect required. However, the docpaths workflow still marks the change as requiring a redirect. This is because, if a page is removed, the workflow only checks the list of configured redirects. This means that it misses the addition of a category index page.

This change adjusts the redirect checking logic to, if a page was deleted in a pull request, check the PR files for additions. If an addition includes the same URL path as the removed file, there is no need for a redirect.

### DIFF
--- a/bot/internal/bot/docpaths.go
+++ b/bot/internal/bot/docpaths.go
@@ -111,6 +111,16 @@ func missingRedirectSources(conf []DocsRedirect, files github.PullRequestFiles) 
 		sources[s.Source] = struct{}{}
 	}
 
+	// Make a map of URL paths for added files so we can check whether a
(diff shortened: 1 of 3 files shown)


fheinecke (approved):

ptgott:
Thanks for the reviews! @fheinecke would you have some time to merge this?

PR #389, opened 2025-08-06
Approval Service Helm Chart

doggydogworld:
This change introduces a Helm Chart for the Approval Service.

It supports Teleport integration through Machine ID to allow the Approval Service to authenticate to Teleport. It does this through the `tbot` chart which is included as a dependency in this chart. It is configured by default to use the IAM Join Method for bots, and the documentation gives examples on extra configuration necessary to setup the `tbot` service to create the identity.

It provides a slightly simpler configuration for the Approval Service, defaulting where necessary to have a more kube native approach. The example documentation has a recommended list of options to configure.


tools/approval-service/helm/approval-service/values.yaml
```diff
@@ -0,0 +1,69 @@
...
+      - name: prod
+        teleport_role: prod-role
+
+appSecrets:
+  webhookSecret: "webhook-secret"
+  installationId: 1234567890 # Should be a numeric value.
+  appId: 1234567890 # Should be a numeric value.
+  # Base64 encoded private key for the GitHub App.
+  # This helps to preserve formatting of the private key.
+  privateKey: "YmFzZTY0LWVuY29kZWQtcHJpdmF0ZS1rZXkK"
```
r0mant, line 49: Should this be here?

fheinecke (commented):
Rather than building and maintaining a Helm chart, can we just use a library chart [like we're doing for Attune](https://github.com/gravitational/cloud-terraform/blob/main/modules/eks-gha/flux-resources/applications/attune/control-plane/helm-release.yaml)?

Is there a tangible benefit to building/maintaining this vs using a common, maintained chart?

doggydogworld:
> Rather than building and maintaining a Helm chart, can we just use a library chart [like we're doing for Attune](https://github.com/gravitational/cloud-terraform/blob/main/modules/eks-gha/flux-resources/applications/attune/control-plane/helm-release.yaml)?
> 
> Is there a tangible benefit to building/maintaining this vs using a common, maintained chart?

Hmm wasn't aware of this option. Looks a bit weird but it does mean we don't have to worry about the lifecycle management of the chart itself. I'll give it a crack later tonight might be a quicker way to get it out.

doggydogworld:
Closing since this was implemented using the Helm library chart approach. Implementation is here: https://github.com/gravitational/cloud-terraform/tree/main/modules/eks-gha/flux-resources/applications/approval-service

PR #393, opened 2025-08-18
docpaths: Fix bug in identifying renames

ptgott:
The docpaths workflow mis-identifies docs path changes as requiring redirects if those changes rename a regular page to a category index page or vice versa. #380 addressed a similar bug, but assumed that the GitHub API would classify changes that include this kind of renaming as additions and deletions.

However, if the GitHub API classifies this kind of change as a rename, the workflow incorrectly looks up a redirect source, even though the URL path for the category index page is identical to the URL for the regular page.

This change addresses the bug by editing the docpaths workflow to, before checking for a redirect source of a renamed file, ensuring that the previous and current URLs are not actually identical.

### DIFF
--- a/bot/internal/bot/docpaths.go
+++ b/bot/internal/bot/docpaths.go
@@ -134,9 +134,16 @@ func missingRedirectSources(conf []DocsRedirect, files github.PullRequestFiles) 
 
 		switch f.Status {
 		case "renamed":
-			p := toURLPath(f.PreviousName)
-			if _, ok := sources[p]; !ok {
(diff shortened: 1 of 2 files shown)


ptgott:
Once we merge this, we can eliminate the otherwise unnecessary redirect introduced here: https://github.com/gravitational/teleport/pull/57865/files#diff-6d9273c3bfa0b82e1f9b00029de2eff3cc18f3f06850bf3301b3586b60f463aaR1638-R1641

PR #400, opened 2025-09-25
Add OS Package Repo Tool 2 (OPRT2)

fheinecke:
## Summary
This is a first pass at a new revision of the [OS Package Repo Tool](https://github.com/gravitational/teleport.e/tree/master/tooling/os-package-repo-tool) (OPRT). The primary purpose of this is to public OS packages via Attune, and is intended to replace practically all of our current tooling around OS package publishing.

## Goals
As discussed internally, this tool will operate under the following assumptions:
* AWS auth is handled outside of program (this is environment specific)
* Teleport auth is handled outside of program (this is environment specific)
* Configuration is provided outside of program (this is environment specific)
* Required binaries (attune CLI, Teleport CLI, others required in the future) will be present in the execution environment

Also discussed internally, this tool is designed to meet the following criteria:
* It must be able to run both locally and in CI/CD pipelines
  * This affects where packages will be pulled from as well as how Attune authentication is handled
* It must run significantly faster than our current process
  * This affects what code needs to run concurrently
* It must support our disaster recovery (rebuilding the entire prod repo from scratch)
  * This affects everything related to supporting a longer runtime, like credential renewal
* It must not require running through a Teleport-backed proxy (e.g. `tsh proxy app attune`)
* It must "nicely" support both being ran as a CLI tool with our current CI/CD pipelines, and should support being used as a library for our replacement of GHA (whatever this ends up looking like)

The only component in this PR that is not built to meet these requirements is config file validation (see schema dir). However I believe that this is important to make it easier to detect problems with changes to our CI/CD pipelines. I also believe that this is less complex and easier to read than re-implementing full config validation and defaulting outside of the jsonschema library.

## Remaining items
This is missing the following pieces that I will add in follow up PRs after this gets an initial review:
* Tests
* A README.md (I will try to add answers to functional questions asked by reviewers to this)
* A GitHub Action and/or reusable workflow to run this in our current pipelines

## Usage

The tool is intended to be used as follows (GHA wrapper to come):

Config file:
```yaml
---
# yaml-language-server: $schema=./config.json
logger:
  level: debug
attune:
  authentication:
    mTLS:
      endpoint: https://attune.ci-cd-cluster.tld
      certificateSource:
        teleport:
          workloadIdentity:
            name: workload-id
packageManagers:
  - apt:
      fileSource:
        s3:
          bucket: some-bucket-name
          path: teleport/tags/
      components:
        # NOTE: this will require updating whenever a new major branch is cut.
        # I'm open to suggestions on making this easier... maybe templating of some kind?
        stable/v18: &major_version_files
          - teleport-(amd|arm)64\.deb
          - teleport-updater\.deb
          - other-packages\.deb
        stable/cloud:
          - teleport-amd64\.deb
        stable/rolling: *major_version_files
      # When adding a new supported OS, update the github.com/gravitational/teleport/blob/master/lib/web/scripts/node-join/install.sh script
      # Otherwise, it will keep using the binary installation instead of the deb repo.
      distros:
        # https://wiki.ubuntu.com/Releases for details
        ubuntu:
          - xenial # 16.04 LTS
          - bionic # 18.04 LTS
          - focal # 20.04 LTS
          - jammy # 22.04 LTS
          - noble # 24.04 LTS
          - plucky # 25.04
        # See https://wiki.debian.org/DebianReleases#Production_Releases for details
        debian:
          - bullseye # 11
          - bookworm # 12
          - trixie # 13
          - forky # 14
```

Run it:
```shell
# Install tools
apt install aws-cli teleport attune

# Platform-specific auth
aws sso login
tsh login

# Run the tool
oprt2 -c /path/to/config.yaml
```


tools/oprt2/pkg/attunehooks/gpg/archive/archiveprovider.go
```diff
@@ -0,0 +1,284 @@
...
+	defer func() {
+		cleanupErr := credDirRoot.Close()
+		if cleanupErr != nil {
+			cleanupErr = fmt.Errorf("failed to close credentials root at %q: %w", ap.credentialsDirectory, err)
+		}
+		err = errors.Join(err, cleanupErr)
+	}()
+
+	// Write all the the archive contents to disk
+	for tarHeader, err := reader.Next(); err != nil; {
```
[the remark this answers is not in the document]
fheinecke: This is wrong. I intentionally handled this by performing all operations on an isolated `os.Root`.

tools/oprt2/pkg/ctxcopy/contextualcopy.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"context"
+	"io"
+)
+
+// Copy is a contextual version of [io.Copy]. If the context is cancelled, calls to
+// read and write functions will error, stopping the copy.
+// This retains all properties of [io.Copy], including support for [io.WriterTo] and
+// [io.ReaderFrom].
+func Copy(ctx context.Context, dst io.Writer, src io.Reader) (int64, error) {
```
zmb3, line 28: This feels like an antipattern to me.

> // If the context is cancelled, calls to
> // read and write functions will error, stopping the copy.

How are you satisfying this? It looks like me like the context is only checked before the read or write operation begins, but as soon as it starts we stop respecting context cancelation and enter a blocking call.

tools/oprt2/pkg/filemanager/s3/s3.go
```diff
@@ -0,0 +1,179 @@
...
+		if err := os.RemoveAll(sfm.workingDirectory); err != nil {
+			cleanupErrs = append(cleanupErrs,
+				fmt.Errorf("failed to remove working directory at %q: %w", sfm.workingDirectory, err))
+		}
+	}
+
+	return errors.Join(cleanupErrs...)
+}
+
+// ListItemsWithPrefix returns a list of items in the storage backend that match the given prefix.
```
zmb3, line 109:
```suggestion
// ListItems returns a list of items in the storage backend.
```

tools/oprt2/pkg/packagemanager/apt/apt.go
```diff
@@ -0,0 +1,224 @@
...
+		// This does not preserve order, but it shouldn't need to
+		for componentName, fileNameMatchers := range components {
+			uniqueMatchers := make(map[string]*regexp.Regexp, len(fileNameMatchers))
+			for _, fileNameMatcher := range fileNameMatchers {
+				uniqueMatchers[fileNameMatcher.String()] = fileNameMatcher
+			}
+
+			if len(uniqueMatchers) != len(fileNameMatchers) {
+				keys := slices.Collect(maps.Keys(uniqueMatchers))
+				slices.Sort(keys)
```
zmb3, line 82:
```suggestion
				keys := slices.Sorted(maps.Keys(uniqueMatchers))
```

r0mant (commented):
@fheinecke Two requests:

- +4K LoC is a lot to review. Please split this PR into several logical parts that are more manageable. Workflows, Makefiles, configs, tool itself can likely be split as well into several logical parts.
- I only glanced through the changeset so far but it all seems very complex. What is contextuacopy.io, io.go, contextlock.go, mutexmap.go, providers, some schema generator? Why is all of this needed, can the implementation be simplified before we spend time on the review? Even Teleport codebase doesn't have this many wrappers over standard library functions and primitives, I'm sure we can do just fine without all of those for the tool that's just supposed to auth with Teleport and call an API to push packages.

fheinecke:
> Please split this PR into several logical parts that are more manageable. Workflows, Makefiles, configs, tool itself can likely be split as well into several logical parts.

Sure, I can do this.

> Why is all of this needed, can the implementation be simplified before we spend time on the review?
> We can always add complexity when needed.

As mentioned in the PR body the complexity of this is stemming from the following business requirements set prior to writing the first line of code for this project:
* It must be able to run both locally and in CI/CD pipelines
  * This affects where packages will be pulled from as well as how Attune authentication is handled
* It must run significantly faster than our current process
  * This affects what code needs to run concurrently
* It must support our disaster recovery (rebuilding the entire prod repo from scratch)
  * This affects everything related to supporting a longer runtime, like credential renewal
* It must not require running through a Teleport-backed proxy (e.g. `tsh proxy app attune`)
  * This necessitates basically building the TLS proxy implementation
* It must "nicely" support both being ran as a CLI tool with our current CI/CD pipelines, and should support being used as a library for our replacement of GHA (whatever this ends up looking like)
  * This necessitates writing modular, reusable code

Yes, all of this could be simplified, but only if we either:
* Cut back on the business asks, or
* Shifted the complexity elsewhere (like fixing the Teleport product issues with uploading lots of large files quickly)

Is there something specific that you'd like me to change here?

> I'm pretty sure even Teleport codebase doesn't have this many wrappers over standard library functions and primitives

Most if not all the contents of these files are here because:
* There are either long-standing open requests for these against the Go stdlib
* There were previously open requests for these against the Go stdlib and they were closed with the recommendation that they be implemented by users outside of the stdlib

I can link to these if you like. Basically all of them are here because the Go encourages/requires the use of a `context.Context` to cancel blocking requests, but the Go standard library is missing support for this in several critical places. Without this:
* Secrets will be leaked in several cases, such as SIGINT/SIGKILL being sent to the program from the environment (CI or local user). 
* The program will not exit gracefully, and will hang in most error cases

PR #402, opened 2025-09-26
Simplify docpaths logic and fix a bug

ptgott:
Currently, the docpaths workflow addresses a corner case in which GitHub marks a file rename as a deletion and an addition. However, there are other possible ways GitHub may classify file renames. For example, if we rename Page A to Page B, and incorporate content from Page C into Page B, the GitHub API may mark Page A as a deletion and Page B as a rename from Page C (rather than an addition).

As a result, in more complex renaming scenarios (e.g., gravitational/teleport#59329), the docpaths workflow can require that the author add a redirect from a file that already exists, i.e., an invalid redirect from one path to the same path.

This change accommodates these more complex rename scenarios. The docpaths workflow checks if the path to check (the previous filename for a renamed file or the filename for a deleted file) corresponds to a file introduced by the PR, without assuming that this would be due to a specific scenario.

Also use an enum for GitHub file statuses to prevent typos in string values from introducing bugs.

### DIFF
--- a/bot/internal/bot/docpaths.go
+++ b/bot/internal/bot/docpaths.go
@@ -111,16 +111,23 @@ func missingRedirectSources(conf []DocsRedirect, files github.PullRequestFiles) 
 		sources[s.Source] = struct{}{}
 	}
 
-	// Make a map of URL paths for added files so we can check whether a
-	// deleted file was actually changed to a category index page or vice
-	// versa.
-	addedPaths := make(map[string]struct{})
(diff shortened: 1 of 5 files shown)


ptgott:
Here's an example of the workflow passing with this change: https://github.com/gravitational/teleport/actions/runs/18046334187/job/51357889251?pr=59329#step:7:16

Otherwise, workflow runs in this PR fail due to a bug.

PR #411, opened 2025-10-22
Add tests for OPRT2 skeleton, APT publishing

fheinecke:
First rounds of tests for OPRT2 based on what has already merged.

### DIFF
--- a/tools/oprt2/README.md
+++ b/tools/oprt2/README.md
@@ -1,0 +1,129 @@
+# OPRT2
+
+The OS Package Repo Tool 2 (OPRT2) is a tool for managing release packages
+within repositories.
+
+## Features
+
+* Repo types
+  * APT
+  * ~~YUM~~ (planned, not implemented yet)
+* Repo managers
+  * [Attune](https://github.com/attunehq/attune)
+    * ~~Authentication sources~~ (planned, not implemented yet)
+      * Token/`Authorization` header
+      * mTLS (requires reverse proxy in front of Attune control plane)
+        * ~~Certificate providers~~ (planned, not implemented yet)
+          * Teleport Machine Workload Identity
+          * Pre-existing PEM-encoded keypairs
+  * Discard (dry-run/no-op)
+* File sources
+  * Local filesystem
+  * S3 bucket
+* ~~GPG sources~~ (planned, not implemented yet)
+  * Pre-existing `.gnupg` directory
+  * Base64-encoded tarball of a `.gnupg` directory
+
+## Example config
+
+```yaml
+---
+# yaml-language-server: $schema=./config.json
+logger:
+  level: debug
+packageManagers:
+  - apt:
+      fileSource:
+        s3:
+          bucket: some-bucket-name  # TODO this will need to support coming from env var because it is enviroment-specific
+          path: teleport/tags/
+      publishingTool:
+        attune:
+          gpg:
+            # Base-64 encoded tarball of `.gnupg` directory, matching what we've done for years
+            # Example command to generate this: `tar --exclude='.#*' --exclude="*~" --exclude="S.*" -czf - ~/.gnupg/ | base64`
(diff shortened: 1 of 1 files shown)


tools/oprt2/pkg/ospackages/managers/apt/apt.go
```diff
@@ -21,29 +21,38 @@ import (
...
 	logger      *slog.Logger
+
+	// Repos is the full config for what APT repos that should be used.
+	// First-level key is repo name (e.g. `ubuntu`, `debian` for Gravitational).
+	// Second-level key is distribution (e.g. `plucky`, `trixie` for Gravitational).
+	// Third-level key is component (e.g. `stable/rolling`, `stable/v18` for Gravitational).
+	// Values are a list of regular expressions that match file paths that should be associated with
+	// with repo/distribution/component combination.
+	// See https://wiki.debian.org/DebianRepository/Format#Overview for details.
+	repos map[string]map[string]map[string][]*regexp.Regexp
```
jimbishopp: Can we define structs for this and the return value of `getRepoFiles`? Maps of maps are difficult to reason about and use. Lots of checking if keys exist or not. Structs are nice because they are self-documenting.

fheinecke: While I 100% agree that this isn't terribly easy to reason around, I don't think that structs are a good fit here. The keys at each level are intentionally variable and are intended to be controlled via user-provided input (such as a config file). Here's an example of what this may look like in YAML notation:

```yaml
ubuntu:
  noble:
    stable/rolling:
      - teleport-(amd64|arm64)\.deb
      - teleport-connect.deb
    stable/cloud:
      - teleport-(amd64|arm64)\.deb
    stable/v18:
      - teleport-(amd64|arm64)\.deb
      - teleport-plugin-(amd64|arm64)\.deb
  bionic:
    stable/v18:
      - teleport-(amd64|arm64)-old-glibc-compatible\.deb
debian:
  trixie:
    stable/rolling:
      - teleport-connect.deb
  sid: # ...
  next-debian-version: # ...
  last-five-debian-versions: # ...
```

The keys at every level are variable and will change as we add/remove support for different distros, versions, and components. This is also branch/major version specific, so we really don't want to have to (continue to) make code changes every time this changes.

I'm open to other options here - any thoughts?

jimbishopp: Okay. After seeing the data structure above I now have even more questions :).

1. Who are the users that will be creating this config file? How often will the configs be different and under what conditions?
2. Why does it look like there is a different pattern for ubuntu than debian? a) `ubuntu/noble/stable/rolling/..`. b) `debian/next-debian-version/...` The pattern I expected was `distribution/component/filename`. It makes it seem like the config file is mixing patterns that define repos/files with repo settings.

We've typically always started by designing the Go data structure. In some conditions where we don't need to support user input, like for [products and entitlements](https://github.com/gravitational/cloud/blob/master/cloud/product/products.go) or the [sheets export config](https://github.com/gravitational/cloud/blob/26bfb32322f1d01771aadf3927934b0155fa7da6/jobs/sheets/export.go#L26), we simply define a Go literal for the config. In others where we need to support an actual config file, we've tried to stick with a config structure that matches the Go data structure so we don't have to write custom encodings (marshal/unmarshal). This makes it easier to test our code because it essentially only includes our own business logic. 

If for whatever reason we do need an extremely flexible config file with different type of patterns, we could use a library like [gjson](https://github.com/tidwall/gjson). That said, I wouldn't worry too much about repetition in the config file to simplify as much as possible. E.g. a file that contains a single line per file like the well documented [debian repo format](https://wiki.debian.org/DebianRepository/Format#Overview), then we wouldn't need to manage regexes at all. But again, I don't have a goo enough understanding of what this file will be used for.

Sorry for the delay. I don't want to slow progress here. I'm looking for a better understanding about the use cases for this so the code we produce is easy to change and test.

Issue #413, opened 2025-10-29
Approval Service sometimes creates duplicate Access Requests for a workflow run


doggydogworld:
## Overview

Currently the Approval Service can sometimes create duplicated Access Requests for a workflow run. This is due to a race where the Approval Service decides whether to create a new Access Request for the workflow run.

The decision to make an Access Request is roughly:
1. Get Deployment Review Event (has `workflow_run_id`)
2. Request list of Access Requests from Teleport
3. If the list already has an Access Request with a matching `workflow_run_id` exit early, otherwise proceed
4. Create the Access Request with the `workflow_run_id` set

The race can happen after the 4th step:

1. Create the Access Request with the `workflow_run_id` set (4th step in previous)
1. Get Deployment Review Event (has `workflow_run_id`) (A new event for same workflow run)
1. Request list of Access Requests from Teleport (while an Access Request with `workflow_run_id` has been created, the list Teleport returns isn't consistent)
1. ... (same as above)

As long as Teleport doesn't return back an Access Request with a matching `workflow_run_id`, duplicates will be created for a workflow run.

PR #414, opened 2025-11-05
adding an event cache to handle burst of requests generating multiple…

aadc-dev:
adding an event cache to handle burst of requests generating multiple access requests. Via a first requests "wins" - subsequent requests within a provided TTL will be rejected and debounced. EventCache implements "clean up on add" semantics, removing expired entries on TryAdd calls.

To fix: https://github.com/gravitational/shared-workflows/issues/413
Cloud-terraform PR adding `event_cache_ttl ` config entry: https://github.com/gravitational/cloud-terraform/pull/7077

### DIFF
--- a/tools/approval-service/internal/service/ttleventcache.go
+++ b/tools/approval-service/internal/service/ttleventcache.go
@@ -80,6 +80,11 @@ func (c *TTLEventCache) TryAdd(id string) bool {
 	return true
 }
 
+// IsBeingProcessed reports whether the given id is currently present in the cache
+//
+// Concurrency and semantics:
+// Acquires c.mu to read the internal map and is safe for concurrent use.
+// It does not modify the cache, extend TTLs, or affect eviction, it only observes presence.
 func (c *TTLEventCache) IsBeingProcessed(id string) bool {
 	c.mu.Lock()
 	defer c.mu.Unlock()
--- a/tools/approval-service/internal/service/ttleventcache_test.go
+++ b/tools/approval-service/internal/service/ttleventcache_test.go
@@ -24,20 +24,19 @@ import (
 	"time"
 )
 
-func (c *TTLEventCache) checkInternalState(t *testing.T, expectedMapLen, expectedListLen int) {
-	t.Helper()
-	c.mu.Lock()
-	defer c.mu.Unlock()
+func checkInternalState(t *testing.T, ec *TTLEventCache, expectedMapLen, expectedListLen int) {
+	ec.mu.Lock()
+	defer ec.mu.Unlock()
 
-	if len(c.cache) != expectedMapLen {
-		t.Errorf("len(c.cache) = %d; want %d", len(c.cache), expectedMapLen)
+	if len(ec.cache) != expectedMapLen {
+		t.Errorf("len(c.cache) = %d; want %d", len(ec.cache), expectedMapLen)
 	}
-	if len(c.expiryList) != expectedListLen {
-		t.Errorf("len(c.expiryList) = %d; want %d", len(c.expiryList), expectedListLen)
+	if len(ec.expiryList) != expectedListLen {
+		t.Errorf("len(c.expiryList) = %d; want %d", len(ec.expiryList), expectedListLen)
 	}
 }
 
-func TestTTLEventCache_WithInvalidTLL_ShouldSetDefaultValue(t *testing.T) {
+func TestTTLEventCache_WithInvalidTTL_ShouldSetDefaultValue(t *testing.T) {
 	defaultTTL := 15 * time.Second
 
 	ecWithZero := NewTTLEventCache(0)
@@ -77,7 +76,7 @@ func TestTTLEventCache_TryAdd_HappyPath(t *testing.T) {
 		if !isSuccessfulFirstTryAdd {
 			t.Errorf("TryAdd(key) returned false on first add, expected true")
 		}
-		ec.checkInternalState(t, 1, 1)
+		checkInternalState(t, ec, 1, 1)
 	})
 }
 
@@ -90,13 +89,13 @@ func TestTTLEventCache_TryAdd_BasicRejectDuplicates(t *testing.T) {
 		if !isSuccessfulFirstTryAdd {
 			t.Errorf("TryAdd(key) returned false on first add, expected true")
 		}
-		ec.checkInternalState(t, 1, 1)
+		checkInternalState(t, ec, 1, 1)
 
 		isSuccessfulSecondTryAdd := ec.TryAdd(keyForTest)
 		if isSuccessfulSecondTryAdd {
 			t.Errorf("TryAdd(keyForTest) returned true on duplicate add, expected false")
 		}
-		ec.checkInternalState(t, 1, 1)
+		checkInternalState(t, ec, 1, 1)
 	})
 }
 
@@ -166,7 +165,7 @@ func TestTTLEventCache_TryAdd_AddAfterExpiry(t *testing.T) {
 		}
 		ec.mu.Unlock()
 
-		ec.checkInternalState(t, 1, 1)
+		checkInternalState(t, ec, 1, 1)
 
 		// try adding original key (should succeed)
 		isSuccessfulRetry := ec.TryAdd(keyABCForTest)
@@ -174,7 +173,7 @@ func TestTTLEventCache_TryAdd_AddAfterExpiry(t *testing.T) {
 			t.Errorf("TryAdd returned false when trying to insert key after after ttl expired, expected true")
 		}
 
-		ec.checkInternalState(t, 2, 2)
+		checkInternalState(t, ec, 2, 2)
 
 	})
 }
@@ -195,12 +194,26 @@ func TestTTLEventCache_EvictionLogic(t *testing.T) {
 			t.Fatal("TryAdd(B) failed")
 		} // "B" expires at T=150ms
 
-		ec.checkInternalState(t, 2, 2)
+		checkInternalState(t, ec, 2, 2)
 
-		// 3. Wait 60ms, time is now T=110ms
+		// 2a. Sleep 49ms (T = 99ms) and verify both A and B are present
+		time.Sleep(49 * time.Millisecond)
+		if ec.Len() != 2 {
+			t.Fatalf("After 99ms: Len() is %d, expected 2 (A, B)", ec.Len())
+		}
+		ec.mu.Lock()
+		if _, ok := ec.cache["A"]; !ok {
+			t.Error("after 99ms: cache does not contain key 'A' (expected present)")
+		}
+		if _, ok := ec.cache["B"]; !ok {
+			t.Error("after 99ms: cache does not contain key 'B' (expected present)")
+		}
+		ec.mu.Unlock()
+
+		// 3. Wait 2ms, time is now T=111ms
 		// "A" (T=100ms) is expired.
 		// "B" (T=150ms) is NOT expired.
-		time.Sleep(60 * time.Millisecond)
+		time.Sleep(1 * time.Millisecond)
 
 		// 4. Add "C", which triggers expire()
 		if !ec.TryAdd("C") {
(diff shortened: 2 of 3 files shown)


fheinecke (approved):
Looks reasonable to me provided that this has been tested end to end

tools/approval-service/internal/service/ttleventcache.go
```diff
@@ -0,0 +1,152 @@
...
+		if ttl > 4*time.Second {
+			return ttl / 4
+		}
+		return 1 * time.Second
+	}
+
+	c := &TTLEventCache{
+		items:           make(map[string]time.Time),
+		ttl:             ttl,
+		cleanupInterval: defaultCleanupInterval(ttl),
```
camscale: this can be simplified to `cleanupInterval: max(ttl / 4, 1*time.Second)`, removing the inline function above.

tools/approval-service/internal/service/ttleventcache.go
```diff
@@ -0,0 +1,152 @@
...
+// evictor runs in a background goroutine and periodically removes expired entries from the cache.
+func (c *TTLEventCache) evictor() {
+	ticker := time.NewTicker(c.cleanupInterval)
+	defer ticker.Stop()
+	defer close(c.done)
+
+	for {
+		select {
+		case <-ticker.C:
+			now := time.Now()
```
camscale: The ticker sends the current time each time it fires, so you can do `case now := <-ticker.C` and remove this line

tools/approval-service/internal/service/ttleventcache.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	if expiry, ok := c.items[eventID]; ok {
+		// unexpired -> reject
+		if now.Before(expiry) {
+			return false
+		}
+
+		// expired -> delete and allow insertion
+		delete(c.items, eventID)
```
camscale: no need to delete here. insertion will just overwrite

tools/approval-service/internal/service/ttleventcache_test.go
```diff
@@ -0,0 +1,140 @@
...
+	const goroutines = 500
+	var wg sync.WaitGroup
+
+	var wins int32
+	var mu sync.Mutex
+
+	// Act
+	for i := 0; i < goroutines; i++ {
+		wg.Go(func() {
+			if ec.TryAdd(eventIDForTest) {
```
camscale: There is no guarantee that all of the calls to `TryAdd()` will happen within 200ms (the ttl). On a slow CI runner with lots of contention, this test could be flaky.

Could you instead use the synctest package to test this behaviour more deterministically?

tools/approval-service/internal/service/ttleventcache_test.go
```diff
@@ -0,0 +1,140 @@
...
+	go func() {
+		done <- ec.TryAdd(eventIDInProgressForTest)
+	}()
+	addResult := <-done
+	if addResult {
+		t.Fatalf("expected concurrent TryAdd to be rejected while cooldown is active")
+	}
+
+	// after TTL elapses, TryAdd should succeed again
+	time.Sleep(ttl + 20*time.Millisecond)
```
camscale: Don't use `time.Sleep` in tests if you are not using [`synctest`](https://pkg.go.dev/testing/synctest)

tools/approval-service/internal/service/ttleventcache_test.go
```diff
@@ -0,0 +1,140 @@
...
+	if !ec.TryAdd(eventIDInProgressForTest) {
+		t.Fatalf("expected first TryAdd to succeed")
+	}
+
+	done := make(chan bool, 1)
+	go func() {
+		done <- ec.TryAdd(eventIDInProgressForTest)
+	}()
+	addResult := <-done
+	if addResult {
```
camscale: On a slow test runner, there's no guarantee that 100ms has not passed in which case a true result here would be valid. This makes the test flaky. Use `synctest` instead to have control over time in tests.

tools/approval-service/internal/service/ttleventcache.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+	}
+
+	cleanup := func() error {
+		_ = c.Stop(context.Background())
+		return nil
+	}
+
+	// start evictor
+	go c.evictor()
```
camscale: There does not really seem to be a need for the evictor to run concurrently. It adds a lot of complexity that is hard to read/comprehend/verify and difficult to test. Instead you can just expire entries as the cache is looked up in `TryAdd()` and do it synchronously. No locking needed, no concurrency, no stopping the cache for cleanup.

It can be made reasonably efficient by keeping an ordered list of expiries as well as the map. On each TryAdd, iterate the expiry slice up to `now`, expiring all entries iterated. Remove them all from the slice. Then just do the map lookup and add an entry if necessary. No concurrency needed, no channels, no mutexes.

camscale: > No locking needed,

Although I realise now the cache may be called concurrently, so a lock for `TryAdd()` will probably be needed.

tools/approval-service/internal/approvalservice/approvalservice.go
```diff
@@ -109,6 +109,11 @@ func NewFromConfig(ctx context.Context, cfg config.Root, opts ...Opt) (*Service,
 		return nil, fmt.Errorf("creating event processor: %w", err)
 	}
 	a.processor = processor
+	defer func() {
+		if err := processor.Close(); err != nil {
+			a.log.Warn("Failed to close event processor")
```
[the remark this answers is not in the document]
aadc-dev: Right at this point, not much, the Close function for the approval service is meant to do service level cleanup in one location. As of this PR, it calls the Stop function on the cache to stop allowing more entries from being processed, and stops the evictor go routine.

tools/approval-service/internal/service/ttleventcache.go
aadc-dev: Yup a lock on TryAdd would still be needed, and each call to TryAdd will potentially take longer, having do cleanup and keeping track of two data structures to perform deletion.

I am still loving delegating the cleanup to the evictor, to keep the TryAdd fast and simple. I guess we would be moving some complexity from one place to another.
Please let me know what you think and if we feel strongly about doing cleanup on TryCatch.

tools/approval-service/internal/service/ttleventcache_test.go
```diff
@@ -0,0 +1,140 @@
...
+// exactly one caller wins (returns true) and others are rejected within the TTL
+func TestTryAddConcurrency(t *testing.T) {
+	// Arrange
+	ttl := 200 * time.Millisecond
+	ec, cleanupFunc, err := service.MakeTTLEventCache(ttl)
+	if err != nil {
+		t.Fatalf("MakeTTLCache error: %v", err)
+	}
+	defer func() {
+		_ = cleanupFunc
```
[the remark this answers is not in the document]
aadc-dev: yeah good catch.
I will also be removing the cleanup func, will be calling ttl.Stop from the service level Stop function

tools/approval-service/internal/service/ttleventcache.go
camscale: I'm not particularly concerned with the efficiency of this - it is a service that is called so infrequently in the scheme of things that I don't think it matters, and the difference would be negligible anyway. I actually think the concurrent evictor is less efficient overall as it will be a ticker firing every 4 seconds largely doing no work as our release rate of approximately 2 releases per week on average means this service is mostly idle.

I am more concerned with simplicity and maintainability particularly as this service is critical to us getting releases out. And the complexity of expiring entries in-line during a `TryAdd()` seems to be to be a lot less than managing the concurrency of the evictor ensuring that the locking is always correct and the channel shutdown logic is correct. Not just now in this PR, but ongoing as the code evolves.

tools/approval-service/internal/service/ttleventcache_test.go
```diff
@@ -0,0 +1,206 @@
...
+		cleanupTicker := 2 * time.Second
+
+		ec, err := service.NewTTLEventCache(ttl, service.WithCleanupInterval(cleanupTicker))
+		if err != nil {
+			t.Fatalf("MakeTTLCache error: %v", err)
+		}
+
+		// ensure evictor is stopped
+		defer func() {
+			ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
```
camscale: Consider using `t.Context()` instead of `context.Background()` in tests.

tools/approval-service/internal/service/ttleventcache.go
```diff
@@ -0,0 +1,120 @@
...
+//
+// Returns true if the ID was added, or false if the ID is already in the cache and has not expired.
+//
+// Calls to TryAdd will prune any expired entries from the cache before performing the check.
+func (c *TTLEventCache) TryAdd(id string) bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	now := time.Now()
+	c.evictExpired(now)
```
fheinecke: This could potentially cause memory usage to balloon down the road under specific conditions if usage increased or changed. Can you add a comment here flagging this so that if we get hit by this in say three years, it's easier to identify what's going on?

tools/approval-service/internal/service/ttleventcache_test.go
```diff
@@ -0,0 +1,280 @@
...
+		keyABCForTest := "ABC"
+		key123ForTest := "123"
+
+		isSuccessfulKeyOneFirstTry := ec.TryAdd(keyABCForTest)
+		if !isSuccessfulKeyOneFirstTry {
+			t.Errorf("TryAdd(keyOne) returned false on first add, expected true")
+		}
+
+		// wait for TTL to pass (virtual sleep)
+		time.Sleep(ttl + 10*time.Millisecond)
```
camscale: You do not need the +10ms here. You should be able to sleep the exact TTL due to being in a synctest bubble. This makes the testing more precise and you are able to test boundary conditions should they matter.

Same in the other tests.

tools/approval-service/internal/service/ttleventcache_test.go
```diff
@@ -0,0 +1,280 @@
...
+	"sync/atomic"
+	"testing"
+	"testing/synctest"
+	"time"
+)
+
+func (c *TTLEventCache) checkInternalState(t *testing.T, expectedMapLen, expectedListLen int) {
+	t.Helper()
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
camscale: This locking should not be needed. If it is, then the test is non-deterministic as it is obviously changing underneath the test check, meaning it is a flaky test. Instead if you do not take the lock the race detector should identify this as a problem if there is any concurrency of note that affects the tests.

tools/approval-service/internal/service/ttleventcache_test.go
```diff
@@ -0,0 +1,280 @@
...
+
+		var successfulAdds int32
+
+		// Schedule one small step per logical request. The synctest harness runs these
+		// steps under a deterministic scheduler so inter-leavings are reproducible.
+		for i := 0; i < distinctKeys; i++ {
+			key := fmt.Sprintf("key-%d", i)
+			for j := 0; j < requestsPerKey; j++ {
+				if ec.TryAdd(key) {
+					atomic.AddInt32(&successfulAdds, 1)
```
camscale: Is this atomic needed? There doesn't seem to be any concurrency in the test or the unit under test. I'd also rename the "Concurrency" in the test to "Multiple" perhaps since there is no concurrency here.

tools/approval-service/internal/service/ttleventcache_test.go
```diff
@@ -0,0 +1,280 @@
...
+		if !ec.TryAdd("B") {
+			t.Fatal("TryAdd(B) failed")
+		} // "B" expires at T=150ms
+
+		ec.checkInternalState(t, 2, 2)
+
+		// 3. Wait 60ms, time is now T=110ms
+		// "A" (T=100ms) is expired.
+		// "B" (T=150ms) is NOT expired.
+		time.Sleep(60 * time.Millisecond)
```
camscale: Consider sleeping 50ms-1 and checking that both are in the cache, and then sleeping 1 and checking the first expired. This tests the actual boundary and is doable with synctest.

aadc-dev:
Tested this on Staging.
Thanks @doggydogworld

r0mant (approved):

PR #428, opened 2025-12-03
updating Approval Service semantic version to pick up TTL event cache

aadc-dev:
This semantic version update picks up https://github.com/gravitational/shared-workflows/pull/414

Adding .idea to git ignore to avoid intellij / Goland IDEA config files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
 go.work*
+
+# editors
+.idea/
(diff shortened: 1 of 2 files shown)


zmb3 (approved):
The PR title or description should probably mention the approval service, since it wasn't clear to me _what_ semantic version we're updating.

PR #463, opened 2026-03-14
[env-kvstore]: Retrieve values from Secrets Manager

bpdohall:
- Adds capability to retrieve runtime environment values from secrets manager and set environment variables for use in GHA jobs.
- Adds an entry to the job summary indicating success/failure of operations
- Update readme example role permission to align with tag based ABAC.
- Includes a new shared package for interfacing with GitHub Actions runtime environment - currently the new package is symlinked into the tool folder, but will be included via go.mod once merged and tagged).,

Sample successful job: https://github.com/gravitational/bpdohall-platformops-gha/actions/runs/23616322392/job/68784594489

Example job summary output:
image


tools/env-kvstore/github/github.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	f, err := os.OpenFile(summaryFile, os.O_APPEND|os.O_WRONLY, 0644)
+	if err != nil {
+		slog.Error("Error opening GITHUB_STEP_SUMMARY file", "error", err)
+		return
+	}
+	defer func() { _ = f.Close() }()
+
+	if _, err := f.WriteString(output.String()); err != nil {
```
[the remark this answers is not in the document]
bpdohall: I was not sure that I had ownership to replace the contents of the output file. Can confirm the file is unique per workflow step:

[workflow-commands#adding-a-job-summary](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#adding-a-job-summary):
> GITHUB_STEP_SUMMARY is unique for each step in a job

tools/env-kvstore/github/github.go
```diff
@@ -0,0 +1,179 @@
...
+var (
+	summary = summaryReporter{}
+)
+
+func init() {
+	summary = summaryReporter{
+		steps:        []string{},
+		stepStatuses: make(map[string][]StepStatus),
+	}
+}
```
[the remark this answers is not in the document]
bpdohall: https://github.com/gravitational/shared-workflows/pull/463/changes#diff-04c6c959d10b5b137aca72f76cc038cfd27e119791a1b456ecc42a1344846c92R103-R108

tools/env-kvstore/github/github.go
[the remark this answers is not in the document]
bpdohall: restructured so we have a new shared package in `libs/github/actions`

libs/github/actions/actions.go
```diff
@@ -0,0 +1,111 @@
...
+const (
+	GithubEnv         = "GITHUB_ENV"
+	GithubOutput	  = "GITHUB_OUTPUT"
+	GithubState	  	  = "GITHUB_STATE"
+)
+
+// MaskSecretValues takes in a slice of secret values and prints GitHub Actions commands to mask
+// those values from log outputs.
+func MaskSecretValues(secrets []string) {
+	fmt.Println()
```
taraspos, line 22: Is this needed for masking to work, or is this a leftover?

libs/github/actions/actions.go
```diff
@@ -0,0 +1,111 @@
...
+	for _, v := range secrets {
+		if v == "" {
+			continue
+		}
+		fmt.Printf("::add-mask::%s\n", v)
```
taraspos, line 27: Would it make sense to also `add-mask` of base64 encoded value as it's the most common way to extract secrets? I know it won't really protect from extracting values as doing `encDase64(secret+"123)` would work around it, bit might be useful to have just in case.

tools/env-kvstore/actions/summary.go
taraspos, line 1: Instead of linking, it for the purpose of this PR, you can add something like  in `go.mod`

```
replace (
	github.com/gravitational/shared-workflows/libs/github/actions ../../../libs/github/actions
)
```

And later on we can replace it with the versioned github library.

tools/env-kvstore/github/github.go
taraspos, line 1: Is this file still needed or it can be removed in favour of new `github/actions` package now?

taraspos (approved):

libs/github/actions/actions.go
bpdohall: I included printing an empty line before the `::add-mask::` workflow command to avoid outputting when the cursor is not at the start of a new line.

Not sure it's needed, but I did observe some misbehaviour where the `::group::` commands didn't create a collapsing section when there was some debug output immediately before the workflow command.

In this case, I'd rather keep the newline and be confident that the values will never appear as plaintext.

libs/github/actions/actions.go
bpdohall: db0b411

tools/env-kvstore/actions/summary.go
bpdohall: I was able to use a commit ref for now: db0b411

tools/env-kvstore/github/github.go
bpdohall: removed db0b411

jimbishopp (approved):

PR #483, opened 2026-05-27
env-kvstore: collect existing GHA build values for migration

bpdohall:
Adds migration support to `env-kvstore` so the action can accept input of existing GitHub Actions secrets/variables then encrypt and upload the payload to S3.

Changes
- migration config is retrieved from kvstore
- makes `values` paramter optional to allow migration-only config
- track GitHub OIDC claim `workflow` for S3 object naming
- upload encrypted data to S3

Example migration flow:

1. Add `env-kvstore` action as a step in the target repo/workflow you want to migrate.
2. Add migration config to the kvstore variables for target repo (PR to `kvstore` repo / `kvstore-cicd` Spacelift stack).
3. When the `env-kvstore` step runs, it uploads existing runtime secret values (age encrypted) to s3 bucket in kvstore account
4. Manually convert secrets from s3 file into SOPS files
5. Switch `env-kvstore` to the non-migration mode (remove migration config), replace references to GHA secrets/vars
6. Remove the temporary migration bucket


tools/env-kvstore/kvstore/migration.go
```diff
@@ -0,0 +1,115 @@
...
+	ghaVarsContextEnv = "GHA_VARS_CONTEXT"
+	// ghaSecretsContextEnv is the environment variable where the existing GitHub Actions secrets context is stored for migration.
+	ghaSecretsContextEnv = "GHA_SECRETS_CONTEXT"
+)
+
+type MigrationConfig struct {
+	AgePublicKey    string `json:"agePublicKey"`
+	S3Bucket        string `json:"-"`
+	ExistingVars    string `json:"vars"`
+	ExistingSecrets string `json:"secrets"`
```
taraspos: These fields store encrypted string, correct?

Better name would be 
```suggestion
	EncryptedVars    string `json:"vars"`
	EncryptedSecrets string `json:"secrets"`
```

tools/env-kvstore/action.yaml
```diff
@@ -24,8 +24,15 @@ inputs:
...
-    required: true
     description: "Values to retrieve from KVStore and set as environment variables. CSV: environment variable name, value type (variable|secret), override key"
+  existing-gh-secrets:
+    description: "Migration: set to ${{ tojson(secrets) }} to persist existing secrets to S3"
+    default: "{}"
+  existing-gh-variables:
+    description: "Migration: set to ${{ tojson(vars) }} to persist existing variables to S3"
+    default: "{}"
+  tool-version:
+    description: Version of the tool to use
```
taraspos: This doesn't seem to be used anywhere

```suggestion
```

taraspos (approved):

tools/env-kvstore/cmd/s3.go
```diff
@@ -0,0 +1,95 @@
...
+
+	slog.Info("Uploading existing GitHub Actions values to S3.", "bucket", bucket, "key", objectKey, "bytes", len(content))
+
+	s3Client := s3.NewFromConfig(u.awsConfig)
+	_, err = s3Client.PutObject(ctx, &s3.PutObjectInput{
+		Bucket:        aws.String(bucket),
+		Key:           aws.String(objectKey),
+		Body:          bytes.NewReader(content),
+		ContentLength: aws.Int64(int64(len(content))),
+		ContentType:   aws.String("application/json"),
```
taraspos, line 80: Maybe add ACL to be explicitly private?

```suggestion
		ContentType:   aws.String("application/json"),
		ACL:           types.ObjectCannedACLPrivate
```


https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3@v1.103.3/types#ObjectCannedACL

tools/env-kvstore/kvstore/migration.go
```diff
@@ -0,0 +1,115 @@
...
+		return MigrationConfig{}, fmt.Errorf("invalid age recipient found in kvstore for migration: %w", err)
+	}
+
+	ghaVarsContext := os.Getenv(ghaVarsContextEnv)
+	ghaSecretsContext := os.Getenv(ghaSecretsContextEnv)
+	if ghaVarsContext == "{}" && ghaSecretsContext == "{}" {
+		return MigrationConfig{}, NewSkipMigrationError(
+			fmt.Sprintf("missing GitHub Actions context for migration: ensure %s and %s environment variables are set", ghaVarsContextEnv, ghaSecretsContextEnv),
+		)
+	}
```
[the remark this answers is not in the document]
bpdohall: This is intentional. The default value of the action input variable is "{}". If the value is configured to `tojson(...)` on an empty context, the input variable will be "{}". 

An empty string indicates a mistake in the supplied input variable and should be passed through.

tools/env-kvstore/cmd/s3.go
bpdohall: Including an ACL is unneccessary:
1. The target bucket was created with "Object Ownership = Bucket owner enforced" which means ACLs are ignored. This is the default for new buckets.
2. PutObject requests that lack the ACL header create private objects by default (ref: [PutObject](https://docs.aws.amazon.com/cli/latest/reference/s3api/put-object.html#options))

tools/env-kvstore/kvstore/migration.go
bpdohall: updated in f9bf19e

tools/env-kvstore/cmd/s3.go
```diff
@@ -0,0 +1,95 @@
...
+		return "", fmt.Errorf("missing required GHA claims: enterprise=%s, repository=%s, run_id=%s, workflow=%s", ghaClaims.Enterprise, ghaClaims.Repository, ghaClaims.RunID, ghaClaims.Workflow)
+	}
+	env := ghaClaims.Environment
+	if env == "" {
+		env = "__noenv__"
+	}
+	return fmt.Sprintf("%s/%s/%s/%s/%s.json", ghaClaims.Enterprise, ghaClaims.Repository, ghaClaims.Workflow, env, ghaClaims.RunID), nil
+}
+
+func (u *migrationUploader) Upload(ctx context.Context) error {
```
evanfreed, line 45: nit: would read more naturally checking `HasMigrationConfig()` up front with an early return .  This keeps the happy path unindented (see [indent error flow](https://go.dev/wiki/CodeReviewComments#indent-error-flow)) and lets `GetMigrationConfig` error mean only "configured but failed."

Example: 
```go
func (u *migrationUploader) Upload(ctx context.Context) error {
      if !u.migrationConfig.HasMigrationConfig() {
          migrationSummary("Migration config not present; skipping collection of existing GitHub Actions values.", actions.SummaryResultInfo)
          return nil
      }

      migrationConfig, err := u.migrationConfig.GetMigrationConfig()
      if err != nil {
          var skipErr kvstore.SkipMigrationError
          if errors.As(err, &skipErr) {
              migrationSummary(skipErr.Error(), actions.SummaryResultWarning)
              return nil
          }
          return fmt.Errorf("error getting migration configuration: %w", err)
      }

```

evanfreed (approved):
Comment isn't a hard blocker

tools/env-kvstore/cmd/s3.go
bpdohall: that makes sense - ty! 0c60237

jimbishopp (approved):

PR #489, opened 2026-06-24
env-kvstore: warn when env specific values are missing from kvstore

bpdohall:
Currently failing for repositories invoking the action with an `environment` set but the target environment has not been configured in the secrets manager kvstore.

Because we're using ABAC/tag-based access to secrets, there's never going to be a "resource not found" error.

This PR simplifies handling of errors encountered while populating the environment-specific values so they are treated and logged as a warning instead of a failure.


taraspos (approved):

tools/env-kvstore/kvstore/secretsmanager.go
```diff
@@ -255,24 +255,15 @@ func (s SecretsManagerValueProvider) repoOrEnvStoreFromSecretARNs(ctx context.Co
...
-	if envArn != "" && envStore.IsEmpty() && isResourceNotFoundException(err) {
+	// This case is expected when migrating existing GHA workflows and the envrionment store has not yet
+	// been created.
+	if envArn != "" && envStore.IsEmpty() && err != nil {
 		slog.Warn("no environment-specific values found in Secrets Manager, only repo-level values will be available", "environment", s.ghaClaims.Environment, "arn", envArn)
-		err = envStoreNotFoundError{msg: fmt.Sprintf("no environment-specific values found in Secrets Manager for environment %s", s.ghaClaims.Environment)}
-	}
-	if envArn != "" && envStore.IsEmpty() && err == nil {
-		slog.Info("environment-specific values are empty in Secrets Manager, only repo-level values will be available", "environment", s.ghaClaims.Environment, "arn", envArn)
+		err = envStoreError{msg: fmt.Sprintf("no environment-specific values found in Secrets Manager for environment %s", s.ghaClaims.Environment)}
```
rosstimothy: Will an envStoreError error ever be populated with a different message? Should an envStoreError have an environment instead of a msg, and should this fmt.Sprintf be in (envStoreError) Error() instead of here?

rosstimothy (commented):
Can we add some test coverage to prevent regressions?

tools/env-kvstore/kvstore/secretsmanager.go
bpdohall: Revised per your feedback in https://github.com/gravitational/shared-workflows/pull/489/changes/7536402ce15ea3ec5e44b126cbd3c2d37e6636f1 and was able to collapse some logic

Added separate handling for the case where unmarshalling fails. This scenario should fail the job. https://github.com/gravitational/shared-workflows/pull/489/commits/f7d99b2be1e0e89e8d8e1a1dfacb1858555ce7fe

mjsmithnh (approved):

tools/env-kvstore/kvstore/secretsmanager_test.go
```diff
@@ -0,0 +1,204 @@
...
+		valuesConfig: []config.EnvValue{
+			{EnvVar: "MY_SECRET", ValueType: "secret"},
+			{EnvVar: "MY_VAR", ValueType: "variable"},
+		},
+	}
+}
+
+func TestMapStoreFromSecretARN_EmptyARN(t *testing.T) {
+	p := newTestProvider(&mockSecretsManagerClient{})
+	store, err := p.mapStoreFromSecretARN(context.Background(), "")
```
rosstimothy: Here and elsewhere, prefer testing.T.Context() over context.Background().
```suggestion
	store, err := p.mapStoreFromSecretARN(t.Context(), "")
```

tools/env-kvstore/Makefile
```diff
@@ -1,5 +1,5 @@
 TOOL_NAME = env-kvstore
 PACKAGE_PATH = ./cmd
-VERSION = v1.0.1
+VERSION = v1.0.2
```
rosstimothy, line 3: How come we're increasing the version?

bpdohall: I actually need to bump that to v1.0.3.

We invoke this workflow via a `uses` step, which requires  downloadable binaries available from this repo's releases.

```yaml
      - name: Get runtime values from kvstore
        id: kvstore
        uses: gravitational/shared-workflows/tools/env-kvstore@tools/env-kvstore/v1.0.2
        with:
          cognito-identity-pool-id: "us-west-2:..."
          cognito-role-arn: "arn:aws:iam::...."
          values: |
            SOME_VAR,variable
            SOME_SECRET,secret
```

tools/env-kvstore/kvstore/secretsmanager_test.go
rosstimothy: Make sure you get _all_ context.Backgrounds in this test.

tools/env-kvstore/Makefile
rosstimothy: Should we increase the version in a separate change?

tools/env-kvstore/kvstore/secretsmanager_test.go
bpdohall: [0ce9fd2](https://github.com/gravitational/shared-workflows/pull/489/commits/0ce9fd23c64dcfa0eca12805f0a5500271bdca4e)

tools/env-kvstore/Makefile
bpdohall: I'm making this change with the intent to use the new version. If it's customary to separate out the version bump, I'm happy to align with whatever folks are doing in this repo.

rosstimothy: It's typically better to separate code change and versioning change. At minimum it would be good to call this type of thing out in the PR description so your reviewers know that it isn't an accidental version bump.

bpdohall: np - I removed the version bump (https://github.com/gravitational/shared-workflows/pull/489/commits/4fcc955805f486ec090f063c7a5865873bad131e) and opened a separate PR: https://github.com/gravitational/shared-workflows/pull/515

tools/env-kvstore/Makefile
```diff
@@ -1,5 +1,5 @@
 TOOL_NAME = env-kvstore
 PACKAGE_PATH = ./cmd
-VERSION = v1.0.2
+VERSION = v1.0.1
```
rosstimothy:
```suggestion
VERSION = v1.0.2
```

bpdohall: https://github.com/gravitational/shared-workflows/pull/489/commits/b00ad0fd715cab8a9a57e2dc527900b52009f28c

rosstimothy (approved):

PR #494, opened 2026-07-01
Extract lib changes from changelog refactor

mjsmithnh:
Lib changes related to https://github.com/gravitational/shared-workflows/pull/493


libs/git/git.go
```diff
@@ -17,78 +17,58 @@ limitations under the License.
...
-// Will work for symbolic references such as tags, HEAD, branches
-func (r *Repo) TimestampForRef(ref string) (time.Time, error) {
-	t, err := r.RunCmd("log", "-1", "--format=%cI", ref)
-	if err != nil {
-		return time.Time{}, trace.Wrap(err, trace.BadParameter("can't get timestamp for ref: %q", ref))
-	}
-	return time.Parse(time.RFC3339, t)
-}
+// prRegex matches the "(#N)" suffix GitHub squash merges append to commit subjects.
+var prRegex = regexp.MustCompile(`\(#(\d+)\)\s*$`)
```
doggydogworld, line 48: There is an edge case where the commit subjects have two PRs (ie `15bc5990461 Server Discovery: fix sudo->sudo-rs migration for ubuntu 25.10+ (#67522) (#67709)`) due to backports. This would only match the latter PR.

TBH I'm not too sure how to handle it as it's likely the changelog is duplicated between the two but it might be worth adding a doc on this just in case this edge case bites us in the future.

doggydogworld (approved):

jimbishopp (approved):

PR #507, opened 2026-07-16
bot: reviewers flag is unused on many workflows

bpdohall:
The `bot` GHA helper only uses the reviewers JSON on 6 of the workflow types.

This PR allows the remaining workflows to start without providing a value for the `reviewers` CLI flag.

To omit a value for workflows where the dependency is "optional" (table below) the flag must be explicitly set to "{}" when running the tool.

Workflow | Dependency |
|---|---|
| `assign` | Hard — `Review.Get()` |
| `check` | Hard — `Review.CheckInternal/External()` |
| `backport` | Optional — `isInternal()` fast path, skip GH API |
| `rfd` | Optional — `GetAdminCheckers()` for `/excluderfd` |
| `exclude-flakes` | Optional — `GetAdminCheckers()` for `/excludeflake` |
| `bloat` | Optional — `GetAdminCheckers()` for `/excludebloat` |

### DIFF
--- a/bot/main.go
+++ b/bot/main.go
@@ -235,6 +235,7 @@ func createBotLocal(ctx context.Context, flags flags) (*bot.Bot, error) {
 // reviewers flag value: assign, bloat, check, exclude-flakes or rfd
 func workflowNeedsReviewers(workflow string) bool {
 	return workflow == "assign" ||
+		workflow == "backport" ||
 		workflow == "bloat" ||
 		workflow == "check" ||
 		workflow == "exclude-flakes" ||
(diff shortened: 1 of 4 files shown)


bot/main.go
```diff
@@ -227,3 +230,14 @@ func createBotLocal(ctx context.Context, flags flags) (*bot.Bot, error) {
...
+
+// workflowRequiresReviewers checks whether the workflow is one that uses the
+// reviewers flag value: assign, bloat, check, exclude-flakes or rfd
+func workflowNeedsReviewers(workflow string) bool {
+	return workflow == "assign" ||
+		workflow == "backport" ||
+		workflow == "bloat" ||
+		workflow == "check" ||
+		workflow == "exclude-flakes" ||
+		workflow == "rfd"
```
jimbishopp: nit
```suggestion
	switch workflow {
	case "assign", "backport", "bloat", "check",  "exclude-flakes", "rfd":
		return true
	}
	return false
```

jimbishopp (approved):

PR #517, opened 2026-08-05
Add exclusion support in preferred reviewers config

hugoShaka:
We have cases where one directory contains both code we want to watch, and generated assets that are often changed and triggers noisy review requests.

This commit adds the ability to exclude paths from preferred reviews by prefixing them with `!`. Excluded paths always take precedence over included ones.

### DIFF
--- a/bot/internal/review/review.go
+++ b/bot/internal/review/review.go
@@ -341,8 +341,8 @@ func (r *Assignments) getPreferredReviewers(teamReviewers map[string]Reviewer, s
 	// To avoid assigning too many reviewers iterate over paths that we have
 	// preferred reviewers for and see if any of them are among the changeset.
 	coveredPaths := make(map[string]struct{})
-	pathToPreferredReviewers, reviewersToExcludedPath := r.getPreferredReviewersMap(teamReviewers, set)
-	for path, reviewers := range pathToPreferredReviewers {
+	index := r.indexPreferredReviewersMap(teamReviewers, set)
+	for path, reviewers := range index.pathToReviewers {
 		if _, ok := coveredPaths[path]; ok {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


bot/internal/review/review.go
```diff
@@ -361,16 +382,17 @@ func (r *Assignments) getPreferredReviewers(teamReviewers map[string]Reviewer, s
 }
 
 // getAllPreferredReviewers returns a list of reviewers that would be
-// preferrable to review the provided changeset. Includes all preferred
-// reviewers for each file path in the changeset.
+// preferrable to review the provided changeset. Includes 1 preferred
+// reviewer for each file path in the changeset.
```
hugoShaka, line 379: Note: I did not change the behaviour, the godoc was wrong. In the case where we have a single file change, and 3 preferred reviewers, only 1 will be returned.

In most cases it's fine because we have many files in the diff, so most preferred reviewers are pulled.

rosstimothy (approved):

bot/internal/review/review.go
```diff
@@ -379,25 +394,48 @@ func (r *Assignments) getAllPreferredReviewers(reviewers map[string]Reviewer, se
...
+	}
+	return false
+}
+
+// getPreferredReviewersMap builds a map of preferred reviewers for file paths, and a map of excluded file paths for reviewers.
+func (r *Assignments) getPreferredReviewersMap(reviewers map[string]Reviewer, set []string) (map[string][]string, map[string][]string) {
+	// map[path][]reviewers
+	included := make(map[string][]string)
+	// map[reviewer][]excluded-paths
+	excluded := make(map[string][]string)
```
r0mant: Having both return arguments be "map of strings to list of strings" but inverted (one is path -> reviewers, another one is reviewers -> paths) is quite confusing. Can we make it into a proper self-documenting struct instead?

hugoShaka: Done in 9c38c3a530ddae2b01065814dd09c86784c43568 where I introduced an index struct.

r0mant (approved):

